PR #5, opened 2019-10-12
General improvements

eugener:


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -127,15 +129,15 @@ public boolean link(ProcessPaths paths, ProjectConfiguration projectConfiguratio
...
-        linkBuilder.command().add(paths.getAppPath().toString() + "/" + appName);
-        linkBuilder.command().add(linux.toString() + "/launcher.o");
-        linkBuilder.command().add(linux.toString() + "/thread.o");
+        linkBuilder.command().add(paths.getAppPath().resolve(appName).toString());
+        linkBuilder.command().add(linuxPath.resolve("launcher.o").toString());
+        linkBuilder.command().add(linuxPath.resolve("thread.o").toString());
         linkBuilder.command().add(objectFile.toString());
         linkBuilder.command().add("-L" + projectConfiguration.getJavaStaticLibsPath());
-        linkBuilder.command().add("-L"+projectConfiguration.getGraalPath()+"/lib/svm/clibraries/"+target.getOsArch2());// darwin-amd64");
+        linkBuilder.command().add("-L"+ Path.of(projectConfiguration.getGraalPath(),"lib","svm","clibraries",target.getOsArch2()));// darwin-amd64");
```
jperedadnr: Add space after commas.

src/main/java/com/gluonhq/substrate/model/Triplet.java
```diff
@@ -37,6 +39,18 @@
     private String vendor;
     private String os;
 
+    public static Triplet fromCurrentOS() {
```
jperedadnr: move it after the other public `Triplet` constructor?

src/main/java/com/gluonhq/substrate/util/FileDeps.java
```diff
@@ -174,7 +174,7 @@ public static boolean setupDependencies(ProjectConfiguration configuration) thro
     }
 
     private static String getChecksumFile(Path unpacked, String name, String osArch) {
-        String answer = unpacked.getParent().toString()+ File.separator + name+"-" + osArch + ".md5";
+        String answer = unpacked.getParent().resolve(name+"-" + osArch + ".md5").toString();
```
jperedadnr: `name + "-"`

src/main/java/com/gluonhq/substrate/model/Triplet.java
eugener: The reason it is first is that it is a static factory function, not a constructor

jperedadnr: Makes sense, somehow I thought it was placed in the middle of both constructors, my bad.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
eugener: Fixed. Please look it over as there are few more small tweaks.

src/main/java/com/gluonhq/substrate/util/FileDeps.java
eugener: Fixed.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -127,15 +125,15 @@ public boolean link(ProcessPaths paths, ProjectConfiguration projectConfiguratio
                     +gvmPath.toString());
         }
         ProcessBuilder linkBuilder = new ProcessBuilder("gcc");
-        Path linux = gvmPath.resolve(appName);
+        Path linuxPath = gvmPath.resolve(appName);
```
johanvos: That name is wrong anyhow, as it is nothing linux specific. I think it should be changed to "appPath" or so?

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -148,32 +146,29 @@ public boolean link(ProcessPaths paths, ProjectConfiguration projectConfiguratio
         linkBuilder.command().addAll(getTargetSpecificLinkFlags());
         linkBuilder.redirectErrorStream(true);
         Process compileProcess = linkBuilder.start();
-        InputStream inputStream = compileProcess.getInputStream();
         int result = compileProcess.waitFor();
         if (result != 0 ) {
             System.err.println("Linking failed. Details from linking below:");
-            printFromInputStream(inputStream);
+            printFromInputStream(compileProcess.getInputStream());
```
johanvos: I'm ok with this change If it works. But I'm not sure it is possible to call getInputStream() after waitFor() has been called and still have the same inputStream?

src/main/java/com/gluonhq/substrate/util/FileDeps.java
```diff
@@ -169,13 +169,14 @@ public static boolean setupDependencies(ProjectConfiguration configuration) thro
         try (FileInputStream fis = new FileInputStream(new File(nameFile));
              ObjectInputStream ois = new ObjectInputStream(fis)) {
             hashes = (Map<String, String>) ois.readObject();
-        } catch (ClassNotFoundException | IOException e) {}
+        } catch (ClassNotFoundException | IOException e) {
+            e.printStackTrace();
```
johanvos: I think we need to propagate the exception here (original code was wrong).
But that might be better to do in another PR -- it's already good to print the stacktrace instead of ignoring it

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -182,9 +177,67 @@ void printFromInputStream(InputStream inputStream) throws IOException {
...
+        Files.createDirectories(workDir);
+
+        ProcessBuilder processBuilder = new ProcessBuilder("gcc");
+        processBuilder.command().add("-c");
+        if (projectConfiguration.isVerbose()) {
+            processBuilder.command().add("-DGVM_VERBOSE");
+        }
+
+        for( String res: C_RESOURCES ) {
+            FileOps.copyResource("/native/linux/" + res, workDir.resolve(res));
```
jperedadnr: this is linux specific??

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
```diff
@@ -35,70 +35,15 @@
 import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Arrays;
```
jperedadnr: Optimize imports

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
eugener: I'm not aware (and docs do no mention) of such a limitation.  Can revert if it does not work.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
eugener: The reason I did it is maintainability.  The code was identical for both Linux and Mac. One place to fix is much better than two. If MacOS code starts to differ, specialized method can be created.

jperedadnr: So far, same files are used for both linux and mac os. So we probably should rename the folder to `posix` (as Johan suggested). I doubt this will work for other platforms, so this will probably will change soon. For now it is okay.

Issue #26, opened 2019-10-23
Run HelloFX on iOS


jperedadnr:

johanvos:
I think this can be closed now?

PR #37, opened 2019-10-27
#26 Add support for JavaFX on iOS

jperedadnr:


src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -41,33 +41,80 @@
...
+            "com.sun.javafx.font.coretext.CGRect",
+            "com.sun.javafx.font.coretext.CGSize",
+            "com.sun.javafx.font.FontConfigManager$FcCompFont",
+            "com.sun.javafx.font.FontConfigManager$FontConfigFont",
+            "com.sun.javafx.iio.ios.IosImageLoader"
+    );
+
+    private static final List<String> ioslibs = Arrays.asList(
+            "-lpthread", "-lz", "-lstrictmath", "-llibchelper",
+            "-ljava", "-ljvm", "-lnio", "-lzip", "-lnet");
```
johanvos: why is jvm between java and nio? I would assume jvm is the last in row.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -41,33 +41,80 @@
...
-                "-Wl,-framework,UIKit",
-                "-Wl,-framework,CoreGraphics",
-                "-Wl,-framework,CoreText",
-                "-Wl,-framework,OpenGLES",
-                "-Wl,-framework,MobileCoreServices");
+                "-isysroot", getSysroot()));
+        if (useJavaFX) {
+            String javafxSDK = projectConfiguration.getJavafxStaticLibsPath().toString();
+            javafxLibs.forEach(name ->
+                    linkFlags.add("-Wl,-force_load," + javafxSDK + "/lib" + name + ".a"));
```
johanvos: At this point, we are guaranteed that the javafxStaticLibsPath exists, and is not null?
(projectConfiguration.getJavafxStaticLibsPath() always returns a valid path?)

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -41,33 +41,80 @@
...
+                "-Wl,-framework,QuartzCore", "-Wl,-framework,ImageIO"));
+        return linkFlags;
     }
 
     @Override
     List<String> getTargetSpecificCCompileFlags() {
         return Arrays.asList("-xobjective-c",
                 "-arch", getArch(),
-                "-Dsvn.targetArch=" + getArch(),
+                "-Dsvm.targetArch=" + getArch(),
```
johanvos: why do we need this here?

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -76,6 +123,8 @@
         Path llcPath = Path.of(projectConfiguration.getGraalPath(),"bin", "llc");
         return Arrays.asList("-H:CompilerBackend=" + Constants.BACKEND_LLVM,
                 "-H:-SpawnIsolates",
+                "-Dtargetos.name=iOS",
```
johanvos: why is this used?

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -147,4 +150,46 @@ public static Path copyFile(Path source, Path destination)  {
         }
         return destination;
     }
+
+    public static void deleteDirectory(Path start) throws IOException {
```
johanvos: add docs about what will happen if e.g. start does not exist, is null, no permission etc, and when an IOException is thrown.
There are many edge cases here, and rather than reading the code, it is easier to read the contract (e.g. JavaDoc)

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -147,4 +150,46 @@ public static Path copyFile(Path source, Path destination)  {
...
+    public static void copyDirectory(Path source, Path destination) {
+        copyFile(source, destination);
+        if (Files.isDirectory(source)) {
+            try {
+                Files.list(source)
+                        .map(Path::getFileName)
+                        .forEach(fileName ->
+                            copyDirectory(source.resolve(fileName), destination.resolve(fileName)));
+            } catch (IOException e) {
+                Logger.logSevere("Error copying " + source.toString() + ": " + e.getMessage());
```
johanvos: You hide the exception here. The caller is not notified and will assume all is fine.

src/main/java/com/gluonhq/substrate/util/XcodeUtils.java
```diff
@@ -38,6 +38,9 @@ private String getSdkDir(String name) {
         }
     }
 
+    public static final Path XCODE_PRODUCTS_PATH = Paths.get(System.getProperty("user.home")).
+            resolve("Library/Developer/Xcode/DerivedData/GluonSubstrate/Build/Products/");
```
johanvos: hardcode "GluonSubstrate" ?

src/main/java/com/gluonhq/substrate/util/ios/Deploy.java
```diff
@@ -96,4 +103,49 @@ public static boolean install(String app) throws IOException, InterruptedExcepti
...
+        if (Files.exists(XCODE_PRODUCTS_PATH)) {
+            Files.walk(XCODE_PRODUCTS_PATH, 1)
+                    .filter(Objects::nonNull)
+                    .filter(path -> path.getFileName().toString().startsWith(appName))
+                    .forEach(path -> {
+                        try {
+                            Logger.logDebug("Removing older version: " + path.getFileName().toString());
+                            FileOps.deleteDirectory(path);
+                        } catch (IOException e) {
+                            Logger.logSevere("Error removing directory at " + path + ": " + e.getMessage());
```
johanvos: exception is hidden.

src/main/resources/native/ios/thread.m
```diff
@@ -0,0 +1,122 @@
...
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+ #include <stdio.h>
+
+void JVM_StartThread() {
```
johanvos: if you don't add this file, where does linking fail? 
It sounds strange we need all those symbols without usin them.

test-project/helloWorld/build.gradle
```diff
@@ -3,10 +3,14 @@ repositories {
...
+def target = System.getProperty("substrate.target")
+
 dependencies {
-    implementation 'com.googlecode.plist:dd-plist:1.22'
-    implementation 'com.github.jnr:jnr-ffi:2.1.10'
-    implementation 'org.bouncycastle:bcpkix-jdk15on:1.49'
+    if ("ios" == target) {
+        implementation 'com.googlecode.plist:dd-plist:1.22'
+        implementation 'com.github.jnr:jnr-ffi:2.1.10'
+        implementation 'org.bouncycastle:bcpkix-jdk15on:1.49'
```
johanvos: why is the test project dependent on this?

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -41,33 +41,80 @@
...
+            "-lpthread", "-lz", "-lstrictmath", "-llibchelper",
+            "-ljava", "-ljvm", "-lnio", "-lzip", "-lnet");
+
+    private static final List<String> javafxLibs = Arrays.asList(
+            "prism_es2", "glass", "javafx_font", "prism_common", "javafx_iio");
 
     @Override
     List<String> getTargetSpecificLinkFlags(boolean useJavaFX, boolean usePrismSW) {
-        return Arrays.asList("-w", "-fPIC",
+        List<String> linkFlags = new ArrayList<>(Arrays.asList("-w", "-fPIC",
```
eugener: Would just `Arrays.asList` be enough? Why do we need list in the list?

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -41,33 +41,80 @@
...
-                "-Wl,-framework,OpenGLES",
-                "-Wl,-framework,MobileCoreServices");
+                "-isysroot", getSysroot()));
+        if (useJavaFX) {
+            String javafxSDK = projectConfiguration.getJavafxStaticLibsPath().toString();
+            javafxLibs.forEach(name ->
+                    linkFlags.add("-Wl,-force_load," + javafxSDK + "/lib" + name + ".a"));
+        }
+        linkFlags.addAll(ioslibs);
+        linkFlags.addAll(Arrays.asList(
```
eugener: Should the list of fixed strings be a constant?

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -90,10 +139,30 @@ public String getAdditionalSourceFileLocation() {
...
     List<String> getTargetSpecificObjectFiles() throws IOException {
         Path gvmPath = paths.getGvmPath();
         Path objectFile = FileOps.findFile(gvmPath, "llvm.o");
-        return Arrays.asList(objectFile.toAbsolutePath().toString());
+        return Collections.singletonList(objectFile.toAbsolutePath().toString());
+    }
+
+    @Override
+    List<String> getJavaFXReflectionClassList() {
+        List<String> answer = super.getJavaFXReflectionClassList();
```
eugener: Depending on how many times this is executed - it can produce different results I think. Cleaner way might be to add both lists to a new list and return that. Also we seem to do this kind of thing many times, should we think of caching for performance reasons?

src/main/java/com/gluonhq/substrate/util/ios/Deploy.java
```diff
@@ -96,4 +103,49 @@ public static boolean install(String app) throws IOException, InterruptedExcepti
         }
         return false;
     }
+
+    public static void addDebugSymbolInfo(Path appPath, String appName) throws IOException, InterruptedException {
+        Path applicationPath = appPath.resolve(appName + ".app");
+        Path debugSymbolsPath = Path.of(applicationPath.toString() + ".dSYM");
```
eugener: This looks strange... should it be `Path debugSymbolsPath =  applicationPath.resolve(".dSYM")`?

src/main/java/com/gluonhq/substrate/util/ios/Deploy.java
```diff
@@ -96,4 +103,49 @@ public static boolean install(String app) throws IOException, InterruptedExcepti
...
+                        try {
+                            Logger.logDebug("Removing older version: " + path.getFileName().toString());
+                            FileOps.deleteDirectory(path);
+                        } catch (IOException e) {
+                            Logger.logSevere("Error removing directory at " + path + ": " + e.getMessage());
+                        }
+                    });
+        }
+
+        String now = DateTimeFormatter.ofPattern("yyyyMMddHHmmss").format(LocalDateTime.now());
```
eugener: Since `DateTimeFormatter` is thread safe, format `DateTimeFormatter.ofPattern("yyyyMMddHHmmss")` should be a constant

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: ok, moved to the end

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: This is taken care by `FileDeps::setupDependencies` that is called before linking. 
I'll add an explicit verification of paths.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: It was set [here](https://github.com/johanvos/substrate/blob/ioslibs/src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java#L61), but it is not needed, indeed. 
I'll remove it.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: Removed

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
johanvos: I don't think that is needed at this place. More important is that the documentation of FileDeps::setupDependencies makes this very clear, and that it is also proven that that method has been called before this method is called.

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: done

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: I know, but there is a recursive method, and the exception within the forEach can't be propagated. I'll use a regular for loop, unless there is a better way

src/main/java/com/gluonhq/substrate/util/XcodeUtils.java
jperedadnr: done

src/main/java/com/gluonhq/substrate/util/ios/Deploy.java
jperedadnr: Same as above (exception within a forEach).

src/main/resources/native/ios/thread.m
jperedadnr: Error without `thread.m`:

```
Undefined symbols for architecture arm64:
      "_JVM_CountStackFrames", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_CurrentThread", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_DumpThreads", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_GetAllThreads", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_HoldsLock", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_Interrupt", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_IsInterrupted", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_IsThreadAlive", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_ResumeThread", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_SetNativeThreadName", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_SetThreadPriority", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_Sleep", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_StartThread", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_StopThread", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_SuspendThread", referenced from:
          _methods in libjava.a(Thread.o)
      "_JVM_Yield", referenced from:
          _methods in libjava.a(Thread.o)
      "_Java_java_lang_Thread_countStackFrames", referenced from:
          _Thread_countStackFrames_62a502bde129f0997933886334feb845c363ba25 in llvm.o
      "_Java_java_lang_Thread_dumpThreads", referenced from:
          _Thread_dumpThreads_39b1e1fb2d2b0ace213e0ac1ec870e72f3e7d2ac in llvm.o
      "_Java_java_lang_Thread_getThreads", referenced from:
          _Thread_getThreads_891ea1885c7d2cf8e0d23e40e1463524f8c997ae in llvm.o
    ld: symbol(s) not found for architecture arm64
    clang: error: linker command failed with exit code 1 (use -v to see invocation)
    Linking failed
```

test-project/helloWorld/build.gradle
jperedadnr: The test-project doesn't have a dependency on substrate, because tests are run before building the jar. We just add manually the path to substrate/build/classes (and resources), as these are compiled before running tests.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: Because `Arrays.asList` returns a list implementation that doesn't support `addAll`, and therefore `linkFlags.addAll(ioslibs);` will fail.

src/main/resources/native/ios/thread.m
johanvos: So these references come from libjava.a, but why are those methods (e.g. _methods) included in linking? We don't use -Wl,load_all etc, so where would the references come from?
It seems unlikely to me there are indirect calls to Thread.o_methods.

test-project/helloWorld/build.gradle
johanvos: do you mean those 3 lines can be removed?

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: It is not fixed yet. In fact we still need to add a way to let the user include more frameworks (Apple's or third party ones). Anyway, this could be part of a static default list.

src/main/java/com/gluonhq/substrate/util/ios/Deploy.java
jperedadnr: No, that will produce a different result. We need to create `appName.app.dSYM` directory.

src/main/java/com/gluonhq/substrate/util/ios/Deploy.java
jperedadnr: Ok

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: I'm not following?
`super` returns already a static list `javafxReflectionClassList`, and we combine it with the platform list (`javafxReflectionIosClassList` in this case) into the resulting list.

test-project/helloWorld/build.gradle
jperedadnr: I didn't say that, quite the contrary: Since we don't have a dependency on substrate, its transitive dependencies aren't resolved, and that's why we have to add them.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
eugener: I thought the original list is updated and then has all the new elements. if this method is called more then once - it will be updated multiple times.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
eugener: I missed that :)

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: Yes, you are right, the original list is updated and a second call will increase it with duplicated items. I'll fix this.

jperedadnr: It should be fixed now

eugener: I think I saw many places like that before :)

test-project/build.gradle
```diff
@@ -26,6 +26,43 @@ def setupDependencies(String platform) {
...
+    if (!javafxSdk.exists()) {
+        javafxSdk.mkdirs()
+    }
+    def version = "14-ea+gvm1"
+    def name
+    if ("linux" == "$platform") {
+        name = "linux-x86_64"
+    } else if ("mac" == "$platform") {
+        if ("ios" == "$target") {
+            name = "ios-arm64"
```
johanvos: This looks extremely brittle. Can we use the Constants?

jperedadnr: Indeed, it should be fixed with #56

Issue #43, opened 2019-10-29
Add HelloWorld ios on travis


johanvos:
We should be careful with travis resources, but running HelloWorld/ios should work in travis.

PR #46, opened 2019-10-29
#43 Test HelloWorld on Travis, without signing.

jperedadnr:
Added IosConfiguration (WIP)

### DIFF
--- a/src/main/java/com/gluonhq/substrate/model/IosConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/IosConfiguration.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright (c) 2019, Gluon
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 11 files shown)


src/main/java/com/gluonhq/substrate/model/IosConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.gluonhq.substrate.model;
+
+import java.util.List;
+
+public class IosConfiguration {
```
johanvos: Might be better to cal this IosSigningConfiguration or so? It only deals with a specific part of configuration.

jperedadnr: It is for signing (that also includes setting given provisioning/identities), but also for including frameworks (iOS or third party)

src/main/java/com/gluonhq/substrate/model/IosConfiguration.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public void setSimulatorDevice(String simulatorDevice) {
+        this.simulatorDevice = simulatorDevice;
+    }
+
+    public String getSimulatorDevice() {
+        return simulatorDevice;
+    }
+
+    public List<String> getFrameworks() {
```
johanvos: I don't think this belongs here. 
What's the difference with additional link parameters?

src/main/java/com/gluonhq/substrate/model/IosConfiguration.java
johanvos: As I stated in the other comment, I don't think frameworks need to be managed here. That is a generic component (libraries)

src/main/java/com/gluonhq/substrate/model/IosConfiguration.java
jperedadnr: You mean frameworks can be also added to Mac?

johanvos: even more general, why would we treat frameworks different from libraries? Is there another use for them apart then providing them to the linker?

Issue #48, opened 2019-10-30
Verify output of test-projects


jperedadnr:
Each test produces different output, but all have in common the use of substrate, so we could verify:
 - build folder creation (autoclient, gym, log, ...)
 - files creation (.o, app, ...) 
 - output (in case of HelloWorld)
...

PR #49, opened 2019-10-30
[WIP] #48 Verify HelloWorld output

jperedadnr:


src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -78,7 +79,8 @@ public static void main(String[] args) throws Exception {
         config.setUsePrismSW(usePrismSW);
         config.getIosSigningConfiguration().setSkipSigning(skipSigning);
 
-        TargetConfiguration targetConfiguration = getTargetConfiguration(targetTriplet);
+        TargetConfiguration targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet),
```
johanvos: Looks good, we probably need more similar constructs that would fail-fast.

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -116,9 +118,13 @@ public static void main(String[] args) throws Exception {
         }
         System.err.println("Running...");
         if (expected != null) {
-            InputStream is = targetConfiguration.run(paths.getAppPath(), appName);
-            // TODO: compare expected and actual output
-
+            String response = targetConfiguration.run(paths.getAppPath(), appName);
```
johanvos: What if the output contains an (almost) infinite number of lines? In case the result is an inputstream, there would be no issues as long as the input is not read, but in case of a String, it will use memory.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -263,19 +265,29 @@ public boolean compileAdditionalSources()
...
-    public InputStream run(Path appPath, String appName) throws IOException {
-        Process runProcess = startAppProcess(appPath,appName);
-        return runProcess.getInputStream();
+    public String run(Path appPath, String appName) throws IOException, InterruptedException {
+        Path app = Objects.requireNonNull(appPath, "Application path can't be null")
+                .resolve(Objects.requireNonNull(appName, "Application name can't be null"));
+        if (!Files.exists(app)) {
+            throw new IOException("Application not found at path " + app.toString());
+        }
+        ProcessRunner runner = new ProcessRunner(app.toString());
```
johanvos: Can you file an issue for creating docs + tests for ProcessRunner? From the method names and signature, it is not clear what they are doing. You currently need to investigate the code to find out, and those methods can have serious impacts on performance, hence developers want to know what they do.

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
jperedadnr: We use a buffer to read the input stream. When the process ends, we return the last line read.
Do you suggest we put a limit in the buffer size and only store a given number of lines?

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: Done

PR #59, opened 2019-11-01
Replace JNI lists with json files

jperedadnr:


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -379,21 +382,25 @@ private Path createJNIConfig(String suffix) throws IOException {
...
-                }
+            for (String javaFile : getJNIClassList(projectConfiguration.isUseJavaFX(), projectConfiguration.isUsePrismSW())) {
+                bw.write(",\n");
+                Files.lines(Paths.get(AbstractTargetConfiguration.class.getResource("/config/" + javaFile).toURI()))
+                        .filter(line -> !line.startsWith("[") && !line.startsWith("]"))
+                        .forEach(line -> {
+                            try {
+                                bw.write(line + "\n");
+                            } catch (IOException e) {
+                                e.printStackTrace();
```
johanvos: probably should be fatal. If writing the JNI json file fails, there will be exceptions later (worst-time at runtime)

Issue #70, opened 2019-11-09
Use reflection files


jperedadnr:
Use reflection json files for each target, and include the methods required only.
Split files, that can be combined as necessary.

PR #71, opened 2019-11-09
#70 Use reflection json files

jperedadnr:


src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
```diff
@@ -80,22 +80,7 @@ String getAdditionalSourceFileLocation() {
...
-            "com.sun.glass.ui.mac.MacView",
-            "com.sun.glass.ui.mac.MacPlatformFactory",
-            "com.sun.glass.ui.mac.MacGestureSupport",
-            "com.sun.glass.ui.mac.MacMenuBarDelegate",
-            "com.sun.glass.ui.mac.MacCommonDialogs",
-            "com.sun.glass.ui.mac.MacFileNSURL",
-            "com.sun.javafx.font.coretext.CTFactory"
-    );
-
+    private static final String javafxReflectionMacClassList = "reflectionconfig-javafx-x86_64-darwin.json";
```
johanvos: better to create this with the Constants for x86_64 and darwin (e.g. in case we move to AMD64 instead of x86-64)

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -133,7 +119,7 @@ public String getAdditionalSourceFileLocation() {
     @Override
     List<String> getJavaFXReflectionClassList() {
         List<String> answer = new ArrayList<>(super.getJavaFXReflectionClassList());
-        answer.addAll(javafxReflectionIosClassList);
+        answer.add(javafxReflectionIosClassList);
```
johanvos: why not answer.add("jniconfig-javax-"+configuration.getArchOs()+".json) on the AbstractTargetConfiguration (with an exists check)?

test-project/helloFX/build.gradle
```diff
@@ -49,9 +45,11 @@ task runScript(type: Exec) {
     def argsList = ["${java_home}/bin/java", '-cp', substratePath,
                     "-Dimagecp=${javafx_path}/javafx.base.jar:${javafx_path}/javafx.graphics.jar:${javafx_path}/javafx.controls.jar:" +
                             "${project.sourceSets.main.runtimeClasspath.filter{!it.name.contains('javafx')}.asPath}",
-                    "-Dgraalvm=$graal_path", "-Dprism.sw=true"]
+                    "-Dgraalvm=$graal_path"]
     if (target != null) {
         argsList += ["-DtargetProfile=$target"]
+    } else {
+        argsList += ["-Dprism.sw=true"]
```
johanvos: if target == null sw rendering needs to be used? 
That sounds unrelated to each other.

jperedadnr: This is for running HelloFX on travis/jenkins without a display. Since we don't run on iOS (target != null), it is an easy way to add that condition.

johanvos: That is very brittle. 
What you want to say is that if the "target" is "ios", you want to specify -DtargetProfile=ios.
If the target is not ios (e.g. it is not set, or it is "linux" or it is "armv6hf" or so, you need software rendering.

But I think what you really want to say is that in case of crosscompilation, don't bother about displaying.
In case of crosscompilation, specify the "target" as a profile.

src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
jperedadnr: Makes sense, I've refactored it.

Issue #77, opened 2019-11-11
Run tests with substrate.jar


jperedadnr:
And its transitive dependencies, avoid adding unnecessary dependencies from the test project to substrate class path

PR #78, opened 2019-11-11
#77 Publish locally and add only substrate and transitive dependencie…

jperedadnr:
…s to classpath

For now it has a hard dependency on the snapshot version

### DIFF
--- a/test-project/helloFX/build.gradle
+++ b/test-project/helloFX/build.gradle
@@ -10,11 +10,11 @@ repositories {
 def target = System.getProperty("substrate.target")
 
 configurations {
-    provided
+    substrate
 }
 
 dependencies {
-    provided "com.gluonhq:substrate:0.0.2-SNAPSHOT"
+    substrate "com.gluonhq:substrate:0.0.2-SNAPSHOT"
 }
 
 javafx {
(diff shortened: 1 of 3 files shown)


test-project/helloFX/build.gradle
```diff
@@ -3,17 +3,18 @@ plugins {
...
 
 repositories {
+    mavenLocal()
     mavenCentral()
 }
 
 def target = System.getProperty("substrate.target")
 
+configurations {
+    provided
```
tiainen: `substrate` might be a better name?

test-project/helloFX/build.gradle
```diff
@@ -3,17 +3,18 @@ plugins {
...
+}
+
 dependencies {
-    if ("ios" == target) {
-        implementation 'com.googlecode.plist:dd-plist:1.22'
-        implementation 'com.github.jnr:jnr-ffi:2.1.10'
-        implementation 'org.bouncycastle:bcpkix-jdk15on:1.49'
-    }
+    provided "com.gluonhq:substrate:0.0.2-SNAPSHOT"
```
tiainen: Are the other dependencies no longer needed on the classpath?

test-project/helloFX/build.gradle
jperedadnr: ok

test-project/helloFX/build.gradle
jperedadnr: These are transitive dependencies, and therefore, they are included in the classpath

tiainen: True.

Issue #40, opened 2019-10-28
Add HelloGluon test


jperedadnr:
Provide support for Gluon Attach in Substrate

PR #79, opened 2019-11-12
#40 Add HelloGluon test

jperedadnr:
Also extends the use of reflectionList, jniList for user custom classes


src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
```diff
@@ -53,6 +53,7 @@ String getAdditionalSourceFileLocation() {
                 libPath + "libprism_es2.a", libPath + "libglass.a",
                 libPath + "libjavafx_font.a", libPath + "libjavafx_iio.a"));
         if (usePrismSW) {
+            answer.add(libPath + "libdecora_sse.a");
```
johanvos: why is this needed? Who calls it?

test-project/gradlew
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env sh
```
johanvos: is this intentionally committed?

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -405,11 +411,23 @@ private Path createJNIConfig(String suffix) throws IOException {
...
             bw.write("]");
         }
         return jniPath;
     }
 
+    private static void writeEntry(BufferedWriter bw, String javaClass) throws IOException {
+        writeEntry(bw, javaClass, false);
+    }
+
+    private static void writeEntry(BufferedWriter bw, String javaClass, boolean exclude) throws IOException {
```
johanvos: we need better names or docs. e.g. what is "exclude"?

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
jperedadnr: In theory the native library is [loaded](https://github.com/openjdk/jfx/blob/22d4343fe8563c2931910b98e8f18c6fd4a48f05/modules/javafx.graphics/src/main/java/com/sun/scenario/effect/impl/sw/sse/SSERendererDelegate.java#L41) from `SSERendererDelegate`.

test-project/gradlew
jperedadnr: Yes, I added a commit with this, as it helps to run locally any test-project.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: The `exclude` parameter might not be used at all, since we are adding concise methods now in the jni/reflection files. Since this is still work in progress, we'll have probably to refactor it later on anyway.

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
jperedadnr: However I'm not sure if it required (at least for the small number of effects we are using now), so probably this can be removed from here.

Issue #91, opened 2019-11-17
Add AllChartSets option


jperedadnr:
We need to enable `"-H:+AddAllCharsets"` to the native image call, otherwise we will get runtime failures like:
```
Caused by: java.nio.charset.UnsupportedCharsetException: UTF-32LE
        at java.nio.charset.Charset.forName(Charset.java:529)
 ```

PR #97, opened 2019-11-18
#91 Add all charsets flag

jperedadnr:


johanvos:
Looks ok, but I wonder what the impact is on the footprint. I'll create a follow-up issue for this.

Issue #92, opened 2019-11-17
Enable URL protocols


jperedadnr:
We need to add `-H:EnableURLProtocols=http,https` to native image.

That will probably require adding some static libraries (`"-lextnext", "-lj2pkcs11", "-lsunec"`) as well.

PR #99, opened 2019-11-18
#92 Add support for URL protocols

jperedadnr:


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -538,6 +532,12 @@ String getAppPath(String appName) {
         return paths.getAppPath().resolve(appName).toString();
     }
 
+    List<String> getCommonLinkLibraries() {
+        return Arrays.asList("-ljava", "-lnio", "-lzip", "-lnet",
+                "-ljvm", "-lstrictmath", "-lz", "-ldl",
+                "-lj2pkcs11", "-lsunec", "-lstdc++");
```
johanvos: remove stdc++ (fails on android)

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
```diff
@@ -61,9 +63,16 @@ String getAdditionalSourceFileLocation() {
         return answer;
     }
 
+    @Override
```
johanvos: semantically, it sounds strange that the "common" libraries can be overridden by platform-specific implementations.
Unless you refer to "common" as for all applications?

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
```diff
@@ -61,9 +63,16 @@ String getAdditionalSourceFileLocation() {
...
+    List<String> getCommonLinkLibraries() {
+        List<String> defaultLinkFlags = new ArrayList<>(super.getCommonLinkLibraries());
+        defaultLinkFlags.add("-lextnet");
+        return defaultLinkFlags;
+    }
+
     private static final List<String> macoslibs = Arrays.asList("-lffi",
             "-lpthread", "-lz", "-ldl", "-lstrictmath", "-llibchelper",
-            "-ljava", "-lnio", "-lzip", "-lnet", "-ljvm", "-lobjc", "-lstdc++",
+            "-ljava", "-lnio", "-lzip", "-lnet", "-ljvm", "-lobjc",
```
johanvos: that means -ljava etc are in the list twice?

src/main/resources/config/reflectionconfig-java.json
```diff
@@ -14,7 +14,9 @@
     "name":"java.lang.Long"
   },
   {
-    "name":"java.lang.Object"
+    "name":"java.lang.Object",
+    "allDeclaredFields":true,
```
johanvos: I think this has a big impact on footprint?

src/main/resources/config/reflectionconfig-java.json
```diff
@@ -69,4 +87,48 @@
     "name":"java.nio.ByteOrder",
     "methods":[{"name":"nativeOrder","parameterTypes":[] }]
   }
+,
+  {
+    "name":"java.nio.file.attribute.PosixFilePermission",
```
johanvos: This is probably not available on Windows

src/main/resources/config/reflectionconfig-java.json
```diff
@@ -69,4 +87,48 @@
...
+    "fields":[{"name":"seed", "allowUnsafeAccess":true}]
+  }
+,
+  {
+    "name":"com.sun.crypto.provider.HmacSHA1",
+    "methods":[{"name":"<init>","parameterTypes":[] }]
+  }
+,
+  {
+    "name":"com.sun.security.sasl.gsskerb.JdkSASL",
```
johanvos: probably optional.

src/main/resources/config/reflectionconfig-javafx.json
```diff
@@ -908,4 +913,9 @@
     "name":"com.sun.javafx.logging.jfr.JFRPulseLogger",
     "methods":[{"name":"createInstance","parameterTypes":[] }]
   }
+,
+  {
+    "name":"com.sun.security.sasl.gsskerb.JdkSASL",
```
johanvos: why is this in JavaFX?

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: ok

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
jperedadnr: Initially these libraries were part of the AbstractTarget, so every target were using them, that's why "common", but since it can be overridden and modified per target, it might required a different name.

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
jperedadnr: yes, that was a known issue. We might need to do a proper refactoring of all these libraries.

src/main/resources/config/reflectionconfig-java.json
jperedadnr: I took it from the agent. Actually there are no fields in Object, so probably it is not even required?

src/main/resources/config/reflectionconfig-java.json
jperedadnr: I can move it to linux/mac

src/main/resources/config/reflectionconfig-javafx.json
jperedadnr: actually it is duplicated from the java file

src/main/resources/config/reflectionconfig-java.json
jperedadnr: What does optional mean? We don't have a way for the user to choose here.

src/main/resources/config/reflectionconfig-java.json
jperedadnr: However, `java.nio.file.attribute.PosixFilePermission` is a valid class from `java.nio.file.attribute`, that should be available in all platforms. 
For now I'll leave it as is.

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
johanvos: yes, changing the name and adding some basic documentation is useful.
Also, some libraries are really hard fixed to be on all platforms (at least libjava.a is hard needed on all JNI-based platforms) so it might make sense to have those common. Not a showstopper though

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
johanvos: Can you file an issue for this so that we keep track of it?

src/main/resources/config/reflectionconfig-java.json
johanvos: No, but an agent should be able to do. We can tackle this later.

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
jperedadnr: Done

src/main/java/com/gluonhq/substrate/target/DarwinTargetConfiguration.java
jperedadnr: This can be tackled as well as part of #102

Issue #103, opened 2019-11-19
Detect Glisten in the classpath


jperedadnr:
If Glisten is present in the classpath, the related classes required for reflection can be added to the reflection config file automatically. 

There is a big compilation time reduction if these classes are added only with the required methods and not opening all methods/fields.

PR #104, opened 2019-11-19
#103 Detect Glisten in class path, add reflection json file

jperedadnr:
(The changes in this PR imply a follow-up change in the Client maven plugin)


jperedadnr:
Makes sense, we should have a way for any given library to provide its own set of json files (like `META-INF/substrate/config/jniconfig-gluon.json`), and then import and merge all the json files. This should be done in a separated issue.

Issue #107, opened 2019-11-21
Test-project should use substrate version


jperedadnr:
Avoid using a hardcoded version

PR #108, opened 2019-11-21
#107 Set substrate version in properties, use development version

jperedadnr:


test-project/build.gradle
```diff
@@ -70,7 +70,14 @@ subprojects {
     apply plugin: 'application'
 
     ext.substrateVersion = "0.0.3"
-
+    def propFile = file('../../gradle.properties')
```
tiainen: `gradle.properties` is located one directory up, not two? It might be safer to use `file("${project.rootDir}/../gradle.properties")` instead.

jperedadnr: Note that we are running from the subproject tests, like `helloWorld`, so that's why we need to go two directories up.

So that doesn't work, `project.rootDir` will give the subproject file:
`test-project/helloWorld/gradle.properties`

gradle.properties
```diff
@@ -1 +1,2 @@
+substrateVersion = 0.0.4-SNAPSHOT
```
tiainen: Rename the property to `version`. You can then remove it altogether from the `build.gradle` of the substrate project.

jperedadnr: done

test-project/build.gradle
tiainen: `project.rootDir` will give the project directory of the root project, as [stated in the docs](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:rootDir). So it should work.

jperedadnr:
close #107

PR #116, opened 2019-11-22
Run tests with GraalVM 19.3 unless skipGraalvmHome is set

jperedadnr:


test-project/build.gradle
```diff
@@ -7,6 +7,14 @@ def setupGraalDependencies(String platform) {
         throw new GradleException("Platform $platform not supported")
     }
 
+    def skipGraalvmHome = Boolean.getBoolean("skipGraalvmHome")
```
tiainen: I would make this a gradle property (so it will be provided with `-P`). That way it's more clear that this is a gradle only thing and not part of substrate itself?

.travis.yml
```diff
@@ -31,13 +31,24 @@ matrix:
             - build-essential
             - libgtk-3-dev
             - libxtst-dev
+      before_script:
+        - mkdir -p $HOME/graalvm && cd $HOME/graalvm
+        - wget https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.0/graalvm-ce-java11-linux-amd64-19.3.0.tar.gz
+        - tar xf graalvm-ce-java11-linux-amd64-19.3.0.tar.gz && cd $TRAVIS_BUILD_DIR
+        - export GRAALVM_HOME=$HOME/graalvm/graalvm-ce-java11-19.3.0/Contents/Home
```
tiainen: Linux does not have the `/Contents/Home` subfolder structure.

.travis.yml
```diff
@@ -31,13 +31,24 @@ matrix:
...
+        - ${GRAALVM_HOME}/bin/gu install native-imagea
```
tiainen: Remove final `a`.

test-project/build.gradle
jperedadnr: ok, make sense

.travis.yml
jperedadnr: Yes, I realized...

.travis.yml
jperedadnr: fixed

src/test/java/com/gluonhq/substrate/IOSTest.java
```diff
@@ -123,6 +124,7 @@ void helloFXTest() {
                 .withProjectDir(new File("test-project"))
                 .withGradleVersion("5.3")
                 .withArguments(":helloFX:clean", ":helloFX:build",
+                        "-DskipGraalvmHome=true",
```
tiainen: This needs to be passed as a gradle property.

test-project/build.gradle
johanvos: What is the goal of this property?

PR #136, opened 2019-11-25
Refactor for a more object oriented design

eugener:
- SubstrateDispatcher is now used as an instance, which helps to share build root and project configuration for all the compilation steps, which are not not static anymore.
- Target configuration retrieval is now method of the target.
- `canCompileTo` is now a method of the target
- `assetGraal` is now a method of project configuration and is renamed to `canRunNativeImage`
- tests and documentation is adjusted accordingly

### DIFF
--- a/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
+++ b/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
@@ -176,7 +176,7 @@ public class SubstrateDispatcher {
      * @throws Exception
      * @throws IllegalArgumentException when the supplied configuration contains illegal combinations
      */
-     boolean nativeCompile(String classPath) throws Exception {
+     public boolean nativeCompile(String classPath) throws Exception {
          config.canRunNativeImage();
         if (classPath != null) {
             boolean useJavaFX = Stream.of(classPath.split(File.pathSeparator))
@@ -205,7 +205,7 @@ public class SubstrateDispatcher {
         return compile;
     }
 
-    boolean nativeLink() throws IOException, InterruptedException {
+    public boolean nativeLink() throws IOException, InterruptedException {
         Triplet targetTriplet  = config.getTargetTriplet();
         TargetConfiguration targetConfiguration = targetTriplet.getConfiguration();
         if (targetConfiguration == null) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/gluonhq/substrate/model/Triplet.java
```diff
@@ -28,6 +28,7 @@
 package com.gluonhq.substrate.model;
 
 import com.gluonhq.substrate.Constants;
+import com.gluonhq.substrate.target.*;
```
johanvos: The only issue I see with this is that Triplet.java will depend on all the specific Configurations. Since Triplet.java is used externally, the configurations are exposed externally as well.
Howevere, there is nothing dangerous in exposing this, I believe.

jperedadnr: Shouldn't we avoid the use of wildcards on imports?

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -200,83 +205,21 @@ public static boolean nativeCompile(Path buildRoot, ProjectConfiguration config,
         return compile;
     }
 
-    public static boolean nativeLink(Path buildRoot, ProjectConfiguration config) throws IOException, InterruptedException {
-        Objects.requireNonNull(config,  "Project configuration can't be null");
+    boolean nativeLink() throws IOException, InterruptedException {
```
jperedadnr: why is it not public? How are we going to call it from the plugins?

src/main/java/com/gluonhq/substrate/model/Triplet.java
eugener: @johanvos You're correct, but the public API uses `TargetConfiguration`, so specific ones are effectively hidden. In fact, I think it can be wise to move `TargetConfiguration` out fo that package, then the package does not have to be exported
@jperedadnr this was done by Idea... and I think it is because practically all of the classes from the package are used

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
eugener: My fault. Will be fixed shortly.

src/main/java/com/gluonhq/substrate/model/Triplet.java
jperedadnr: You can configure Idea to use all the imports?

eugener: Probably, but it this particular case it was correct :)

PR #144, opened 2019-11-28
Refactor `FileDeps` to simplify multiple APIs 

eugener:
- `FileDeps` is used as an instance now, which simplifies its API
- File related static methods have been moved from `FileDeps` to `FileOps`
- Target Configurations, use `ProcessPaths` and `ProjectConfiguration` as constructor parameters, instead of passing them to each method
- `SubstrateDispatcher` creates and holds `ProcessPaths` and `TargetConfiguration` for use in mutiple methods, which results in simpler API

### DIFF
--- a/src/main/java/com/gluonhq/substrate/util/FileDeps.java
+++ b/src/main/java/com/gluonhq/substrate/util/FileDeps.java
@@ -36,7 +36,6 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.net.URL;
 import java.nio.channels.Channels;
@@ -46,9 +45,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.attribute.PosixFilePermission;
-import java.security.DigestInputStream;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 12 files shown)


src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -240,4 +232,44 @@ public static void copyDirectory(Path source, Path destination) throws IOExcepti
...
+             ObjectInputStream ois = new ObjectInputStream(fis)) {
+            hashes = (Map<String, String>) ois.readObject();
+        } catch (IOException | ClassNotFoundException e) {
+            Logger.logDebug("Exception trying to get hashmap for "+nameFile+": "+e);
+            return null;
+        }
+        return hashes;
+    }
+
+    static String calculateCheckSum(File file) {
```
johanvos: add some doc, what is this method doing?

eugener: Yes. This might need to move to `FileOps`  too

src/main/java/com/gluonhq/substrate/util/FileDeps.java
```diff
@@ -364,16 +334,16 @@ private static void downloadJavaFXZip(String osarch, Path substratePath, Project
...
-    public static void downloadZip(String urlZip, Path destination) throws IOException {
+    public void downloadZip(String urlZip, Path destination) throws IOException {
         String zip = destination.toAbsolutePath().toString()+".zip";
         Path zipPath = Paths.get(zip);
-        processZip(urlZip, zipPath, destination.getFileName().toString(), null, null);
+        processZip(urlZip, zipPath, destination.getFileName().toString(), null);
     }
 
-    private static void processZip(String urlZip, Path zipPath, String folder, String version, ProjectConfiguration configuration) throws IOException {
+    private void processZip(String urlZip, Path zipPath, String folder, String version) throws IOException {
```
johanvos: add docs about the parameters

eugener: Will leave this alone for now... I think this method can be refactored. Once it's done - will add the docs

src/main/java/com/gluonhq/substrate/util/FileOps.java
eugener: Done

PR #149, opened 2019-11-29
Split project configuration into public and private

eugener:


src/main/java/com/gluonhq/substrate/ProjectConfiguration.java
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Sets the Java static SDK version
+     * @param javaStaticSdkVersion the Java static SDK version
+     */
+    public void setJavaStaticSdkVersion(String javaStaticSdkVersion) {
+        this.javaStaticSdkVersion = javaStaticSdkVersion;
+    }
+
+    public String getJavaStaticSdkVersion() {
+        return Objects.requireNonNull( this.javaStaticSdkVersion, "Java Static SDK version is required" );
```
johanvos: This causes tests to fail, and I think behavior should be as follows:
* if the developer somehow specified a "javastaticsdkversion" (which is highly discouraged, I think we don't even want to expose this), that "version" is returned. 
* else, a Constants value should be returned.

src/main/java/com/gluonhq/substrate/ProjectConfiguration.java
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Sets the JavaFX static SDK version
+     * @param javafxStaticSdkVersion the JavaFX static SDK version
+     */
+    public void setJavafxStaticSdkVersion(String javafxStaticSdkVersion) {
+        this.javafxStaticSdkVersion = javafxStaticSdkVersion;
+    }
+
+    public String getJavafxStaticSdkVersion() {
+        return Objects.requireNonNull( this.javafxStaticSdkVersion, "JavaFX Static SDK version is required" );
```
johanvos: same as above. In general, developers should be discouraged to set properties. The defaults should simply work.

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -67,25 +65,21 @@ public static void main(String[] args) throws Exception {
...
-        ProjectConfiguration config = new ProjectConfiguration();
-        config.setGraalPath(graalVM);
-        config.setMainClassName(mainClass);
-        config.setAppName(appName);
-        config.setJavafxStaticSdkVersion(Constants.DEFAULT_JAVAFX_STATIC_SDK_VERSION);
-        config.setTarget(targetTriplet);
+        ProjectConfiguration publicConfig = new ProjectConfiguration(mainClass);
+        publicConfig.setGraalPath(graalVM);
+        publicConfig.setAppName(appName);
+        publicConfig.setJavafxStaticSdkVersion(Constants.DEFAULT_JAVAFX_STATIC_SDK_VERSION);
```
johanvos: remove this, so that the default is used automatically.
It makes no sense developers have to retrieve the default version and set that, this is what the configuration should hide.

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -67,25 +65,21 @@ public static void main(String[] args) throws Exception {
...
-            config.setReflectionList(Arrays.asList(reflectionList.split(",")));
-        }
-        if (jniList != null && !jniList.trim().isEmpty()) {
-            config.setJniList(Arrays.asList(jniList.split(",")));
-        }
-        if (bundlesList != null && !bundlesList.trim().isEmpty()) {
-            config.setBundlesList(Arrays.asList(bundlesList.split(",")));
-        }
+        config.setJavaStaticLibs(staticLibs);
+        config.setJavaFXStaticSDK(staticJavaFXSDK);
```
johanvos: should not be set by a developer

src/main/java/com/gluonhq/substrate/ProjectConfiguration.java
johanvos: The logic should probably be in the PrivateProjectConfiguration. The ProjectConfiguration itself is simply a value holder for e.g. plugins to advise on configuration

eugener: Fixed

src/main/java/com/gluonhq/substrate/ProjectConfiguration.java
eugener: Fixed

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
eugener: Fixed

eugener: Fixed

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
eugener: Which one? Whats the default behavior for this?

johanvos: The InternalConfiguration

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -28,20 +28,19 @@
...
-import com.gluonhq.substrate.util.Logger;
 
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.Arrays;
-import java.util.Objects;
-import java.util.Optional;
+import java.util.*;
```
jperedadnr: Make sure you set IntelliJ not to use wildcards in imports (set Preferences -> Editor > Code Style > Java > Imports tab set Class count to use import with '*' -> 99)

src/main/java/com/gluonhq/substrate/model/PrivateProjectConfiguration.java
```diff
@@ -50,25 +49,19 @@
  * If this method has not been called, getJavaStaticLibsPath() will return the default location, taking into account
  * the value of javaStaticSdkVersion. If that value is not set, the default value is used.
  */
-public class ProjectConfiguration {
+public class PrivateProjectConfiguration {
```
jperedadnr: I don't like the class name using `Private`, maybe `InternalProjectConfiguration`?

src/main/java/com/gluonhq/substrate/model/PrivateProjectConfiguration.java
```diff
@@ -78,19 +71,40 @@
...
-    private String appName;
-    private String mainClassName;
-
     private IosSigningConfiguration iosSigningConfiguration = new IosSigningConfiguration();
 
-    public ProjectConfiguration() {}
+    private ProjectConfiguration publicConfig;
+
+    /**
+     * Private projects configuration, which includes everythign, including public settings
```
jperedadnr: typo

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
johanvos: I agree with avoiding wildcards (99 being ok, as in some cases there might be static imports of 100+ entities)

src/main/java/com/gluonhq/substrate/ProjectConfiguration.java
```diff
@@ -0,0 +1,246 @@
...
+
+    public ProjectConfiguration( String mainClassName ) {
+        this.mainClassName = Objects.requireNonNull(mainClassName, "Main class name is required")
+                               .contains("/") ?
+                                  mainClassName.substring( mainClassName.indexOf("/") + 1) : mainClassName;
+
+    }
+
+    public Path getGraalPath() {
+        return Path.of( Objects.requireNonNull(this.graalPath, "GraalVM Path is not defined"));
```
johanvos: This delegates the responsibility of detecting/setting GraalVM to the plugin, where it is actually Substrate that should do this. We can do that in a follow-up PR though

eugener: Lets do this in the other PR

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/PrivateProjectConfiguration.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/PrivateProjectConfiguration.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -78,19 +71,40 @@
...
+
+        setUsePrismSW(usePrismSW);
+        getIosSigningConfiguration().setSkipSigning(skipSigning);
 
-    public String getGraalPath() {
-        return this.graalPath;
+        setJavaStaticLibs(System.getProperty("javalibspath")); // this can be safely set even if null. Default will be used in that case
+        setJavaFXStaticSDK(System.getProperty("javafxsdk"));  // this can be safely set even if null. Default will be used in that case
+
+        setReflectionList(splitString(System.getProperty("reflectionlist")));
```
jperedadnr: In `SubstrateDispatcher::main`, `reflectionList` (and other lists), come from a System property, however, when we use the plugin, the plugin itself will create an instance of `ProjectConfiguration`, will call the setters, so in this case we should use here `setReflectionList(config.getReflectionList())`.

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -50,42 +56,20 @@ public static void main(String[] args) throws Exception {
...
-        config.setMainClassName(mainClass);
         config.setAppName(appName);
-        config.setJavafxStaticSdkVersion(Constants.DEFAULT_JAVAFX_STATIC_SDK_VERSION);
         config.setTarget(targetTriplet);
-        config.setUsePrismSW(usePrismSW);
-        config.getIosSigningConfiguration().setSkipSigning(skipSigning);
-        Optional.ofNullable(staticLibs).ifPresent(config::setJavaStaticLibs);
-        Optional.ofNullable(staticJavaFXSDK).ifPresent(config::setJavaFXStaticSDK);
-        if (reflectionList != null && !reflectionList.trim().isEmpty()) {
-            config.setReflectionList(Arrays.asList(reflectionList.split(",")));
```
jperedadnr: Following the previous comment, this should be kept here.

PR #150, opened 2019-11-30
#148 Extract native libraries from jars, iOS target

jperedadnr:


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -80,6 +82,7 @@ String processClassPath(String cp) throws IOException {
     @Override
     public boolean compile(String cp) throws IOException, InterruptedException {
         String classPath = processClassPath(cp);
+        extractNativeLibs(cp);
```
johanvos: why do we need this at compiletime?

jperedadnr: We need to parse the classpath. This could be done in the link phase, but that will require a change in the API to pass the classpath again.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -602,4 +606,68 @@ String getAppPath(String appName) {
...
+    }
+
+    /**
+     * A filter can be used to verify if the native library matches certain
+     * criteria, like being available for a given architecture
+     *
+     * @return a predicate, default is null (no filter applied)
+     */
+    Predicate<Path> getTargetSpecificNativeLibsFilter() {
+        return null;
```
johanvos: would a default that does nothing be more appropriate?

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -307,4 +323,48 @@ public static String calculateCheckSum(File file) {
...
+     *
+     * @param extension the extension of the files in the jar that will be extracted
+     * @param sourceJar the path to the jar that will be inspected
+     * @param target the path of the folder where the files will be extracted
+     * @param filter a predicate that the files in the jar should match.
+     * @throws IOException
+     */
+    public static void extractFilesFromJar(String extension, Path sourceJar, Path target, Predicate<Path> filter) throws IOException {
+        if  (!Files.exists(sourceJar)) {
+            return;
```
johanvos: when will this happen? Doesn't this indicate an exception? Someone asks to extracts native libs from a non-exiting jar?

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -307,4 +323,48 @@ public static String calculateCheckSum(File file) {
...
+        }
+        if  (!Files.exists(target)) {
+            Files.createDirectories(target);
```
johanvos: What if target already exists? How can that happen? In that case, is the processing still relevant?

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: This is the default, a null predicator that let the file pass the filter. We could return `p -> true`, but with null we don't need to evaluate it.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
johanvos: makes sense. I see the extracted libs are stored in a specific path (which is then the contract between compile and link phase). I don't think that is bad (at least I have no better solution).

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: This is just a regular check to bail out if the sourcePath is not a valid path. While that is not expected, this is a method in FileOps that could be used for many different purposes.
I wouldn't throw an exception, but a log can be added.

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: target is the folder where the files will be extracted to. It could exist from a previous run, or if we call several times this function (i.e. from different jars). We only need to create it the first time.

johanvos: But if it exists and it contains wrong data (e.g. from previous runs), wouldn't that lead to issues with wrong symbols? Wouldn't it be better to remove the directory at the beginning of the cycle, create the directory on the first occasion it is needed, and bail if it already exists?

PR #166, opened 2019-12-06
Consolidate classpath operations

eugener:
- Consolidates all class path operations into ClassPath class. 
- Tests
- Docs


gradle.properties
```diff
@@ -1,4 +1,4 @@
 # Substrate Version
-version = 0.0.5-SNAPSHOT
+version = 0.0.5
```
johanvos: keep SNAPSHOT until we release

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,102 @@
...
+ * Represents the classpath
+ * Consolidates all classpath manipulations
+ */
+public class ClassPath {
+
+    private final String classPath;
+
+    /**
+     * Creates the class path
+     * @param classPath standard java classpath, usually delimited with "pathSeparator"
```
johanvos: usually? Does this mean it is not required, or do you mean it can be a single path?

gradle.properties
eugener: Wasn’t me .. it was the merge. Will fix

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public boolean contains( Predicate<String> predicate ) {
+        Objects.requireNonNull(predicate);
+        return asStream().anyMatch( predicate);
+    }
+
+    /**
+     * Returns a list of strings consisting of the elements of this classpath that match
+     * the given predicate.
+     * @param predicate predicate to apply to elements of this classpath
```
johanvos: add "should not be null"

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,102 @@
...
+        Objects.requireNonNull(predicate);
+        return asStream().filter(predicate).collect(Collectors.toList());
+    }
+
+    /**
+     * Returns a List consisting of the results of applying the given
+     * function to the elements of this classpath.
+     *
+     * @param <T> The element type of the resulting List
+     * @param mapper a <a href="package-summary.html#NonInterference">non-interfering</a>,
```
johanvos: ... should not be null

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Single path.. usually is the wrong word I guess

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,102 @@
...
+    public <T> List<T> mapToList( Function<String, T> mapper) {
+        Objects.requireNonNull(mapper);
+        return asStream().map(mapper).collect(Collectors.toList());
+    }
+
+    /**
+     * Returns a String classpath consisting of the results of applying the given
+     * function to the elements of this classpath.
+     *
+     * @param mapper a <a href="package-summary.html#NonInterference">non-interfering</a>,
```
johanvos: ... should not be null.
The reason I think it is important to clearly state this in the doc is that in some cases, a developer might apply an identity transformation. It should be clear that this can not be achieved by passing null

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,102 @@
...
+     */
+    public String mapToString( Function<String, String> mapper) {
+        Objects.requireNonNull(mapper);
+        return asStream().map(mapper).collect(Collectors.joining(File.pathSeparator));
+    }
+
+    /**
+     * Returns a String classpath consisting of the results of applying the given
+     * function to the elements of this classpath. Tries to find javafx libraries and replace them with full path to this library
+     * within provided JavaFX SDK
```
johanvos: There is nothing JavaFX specific to this method. IIUC, this simply replace a set of given jar files with jar files that are found in a given directory, provided they are indeed present in that given directory.

Hence, I think the wording "fx" should be removed from the signature.

gradle.properties
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,101 @@
...
+ * Represents the classpath
+ * Consolidates all classpath manipulations
+ */
+public class ClassPath {
+
+    private final String classPath;
+
+    /**
+     * Creates the class path
+     * @param classPath standard java classpath, delimited with "pathSeparator". Should not be null.
```
jperedadnr: use `{@code File.pathSeparator}` instead of quotes?

src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -0,0 +1,101 @@
...
+        Objects.requireNonNull(predicate);
+        return asStream().filter(predicate).collect(Collectors.toList());
+    }
+
+    /**
+     * Returns a List consisting of the results of applying the given
+     * function to the elements of this classpath.
+     *
+     * @param <T> The element type of the resulting List
+     * @param mapper a <a href="package-summary.html#NonInterference">non-interfering</a>,
```
jperedadnr: Why the link? we don't have `package-summary.html`, do we? Does it refer to the Stream package?

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Done

src/main/java/com/gluonhq/substrate/model/ClassPath.java
eugener: Removed links to simplify docs

eugener:
For some reason this pull request is not built/checked automatically anymore?

jperedadnr:
CI build seems to be running, but when getting into Travis, I get `Under Maintenance`, so maybe we have to wait for a while.

PR #167, opened 2019-12-06
Release from Travis causes split repository issue

abhinayagarwal:
A summary of the problem can be found here: https://github.com/Codearte/gradle-nexus-staging-plugin/issues/76#issuecomment-451615386

This PR also contains a small fix for committing the correct file with version information.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,6 +2,7 @@ plugins {
     id 'java'
     id 'maven-publish'
     id 'signing'
+    id "de.marcphilipp.nexus-publish" version "0.4.0"
 }
 
 group = "com.gluonhq"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -91,26 +92,18 @@ publishing {
...
-        }
-    }
 }
 
 signing {
     required = { gradle.taskGraph.hasTask(publish) && !version.endsWith("SNAPSHOT") }
     sign publishing.publications.maven
 }
+
+nexusPublishing {
```
tiainen: I believe the gradle task for publishing should be changed from `publish` to `publishToSonatype`?
At the following line: https://github.com/gluonhq/substrate/pull/167/files#diff-6e62ddc767174617641f0b68eedf525fR9

abhinayagarwal: We can specifically make a call to `publishToSonatype`. Or, we can keep calling `publish` and it will make a call to `publishMavenJavaPublicationToSonatypeRepository` which does the same task.

PR #175, opened 2019-12-09
Take into account possible naming convention for static jars

jperedadnr:


src/main/java/com/gluonhq/substrate/model/ClassPath.java
```diff
@@ -85,12 +86,24 @@ public String mapToString( Function<String, String> mapper) {
      * @param libNames library names to look for
      * @return the string classpath
      */
-    public String mapWithLibs(Path libsPath, String... libNames ) {
+    public String mapWithLibs(Path libsPath, String... libNames) {
```
eugener: This is a very generic method... I don't think this is a good place for such a concrete change

jperedadnr:
Fixed with new static builds

PR #176, opened 2019-12-09
Refactor processZip, adding tests

jperedadnr:


src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -365,4 +373,127 @@ public static void extractFilesFromJar(String extension, Path sourceJar, Path ta
...
+     * @return a map with the file names and their checksum values
+     * @throws IOException
+     */
+    public static Map<String, String> unzipFile(Path sourceZip, Path targetDir) throws IOException {
+        Objects.requireNonNull(sourceZip);
+        Objects.requireNonNull(targetDir);
+        if (!Files.exists(sourceZip)) {
+            throw new IOException("Error: " + sourceZip + " does not exist");
+        }
+        if (!sourceZip.toString().endsWith(".zip")) {
```
eugener: Does the file have to have that '.zip' extension? Technically it can be any, as long as it is actually a zip archive.

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -365,4 +373,127 @@ public static void extractFilesFromJar(String extension, Path sourceJar, Path ta
...
+            throw new IOException("Error: " + targetDir + " is not a directory");
+        }
+        if (!Files.exists(targetDir)) {
+            Files.createDirectories(targetDir);
+        }
+        Map<String, String> hashes = new HashMap<>();
+        try (ZipInputStream zis = new ZipInputStream(new FileInputStream(sourceZip.toFile()))) {
+            ZipEntry zipEntry = zis.getNextEntry();
+            while (zipEntry != null) {
+                File destFile = new File(targetDir.toFile(), zipEntry.getName());
```
eugener: Probably easier to keep `destFile` as a `Path`

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -365,4 +373,127 @@ public static void extractFilesFromJar(String extension, Path sourceJar, Path ta
...
+                        int len;
+                        while ((len = zis.read(buffer)) > 0) {
+                            fos.write(buffer, 0, len);
+                        }
+                    }
+                    hashes.put(destFile.getName(), calculateCheckSum(destFile));
+                }
+                zipEntry = zis.getNextEntry();
+            }
+            zis.closeEntry();
```
eugener: Closing may never occur if exception is thrown. `Try/finally` or `Try with resources` is needed.

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -365,4 +373,127 @@ public static void extractFilesFromJar(String extension, Path sourceJar, Path ta
...
+     * @param level0 the folder under targetPath, not null, e.g. foo1
+     * @param level1 a folder under level0, it can be null, e.g. foo2
+     * @param level2 a folder under level1, it can be null, e.g. foo3
+     *              The zip file will be downloaded into /opt/bar/foo.zip
+     *              The contents of this zip file will be installed into
+     *              /opt/bar/foo1/foo2/foo3, and also the file
+     *               /opt/bar/foo1/foo2/foo3/foo1-foo3.md5 will be created
+     * @throws IOException
+     */
+    public static void processZip(String zipUrl, Path targetPath, String zipName,
```
eugener: Should the method be named `downloadAndUnzip`?
I might make sense to rethink parameter names:
- `zipUrl -> sourceUrl`
- `targetPath -> destPath` 

Also... does it make sense to define `level` parameters as varargs?

src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -365,4 +373,127 @@ public static void extractFilesFromJar(String extension, Path sourceJar, Path ta
...
+        if (level2 != null) zipDir = zipDir.resolve(level2);
+
+        Files.createDirectories(zipDir);
+
+        // 3. Extract zip from zipPath into zipDir
+        Map<String, String> hashes = FileOps.unzipFile(zipPath, zipDir);
+
+        // 4. Write hashes file into zipDir
+        try (FileOutputStream fos =
+                     new FileOutputStream(zipDir.toString() + File.separator + md5name);
```
eugener: Much safer to use `Path` API for the name to avoid using `Path.separator` IMO

src/test/java/com/gluonhq/substrate/util/FileOpsTests.java
```diff
@@ -108,4 +112,87 @@ void copyNullStreamDestination() throws IOException {
...
+    void downloadNullPath() {
+        assertThrows(NullPointerException.class, () -> FileOps.downloadFile(new URL(TEST_URL), null));
+    }
+
+    @Test
+    void downloadFile() throws IOException {
+        Path testPath = getTempDir().resolve("test.txt");
+        assertDoesNotThrow(() -> FileOps.downloadFile(new URL(TEST_URL + "test.txt"), testPath));
+        assertTrue(Files.exists(testPath));
+        assertEquals("some dummy tekst test", Files.readAllLines(testPath).get(0));
```
eugener: Spelling `tekst -> text`?

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: True, removing this condition, even if the file (with any extension) is not a valid zip, won't fail: the `zipEntry` will be null and it won't process it. As a result, the hash map will be empty. Not sure if we want this to happen though?

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: ok

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: Not sure if I understand it correctly, but the whole zip process is within a try with resources.

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: Ok, done

src/main/java/com/gluonhq/substrate/util/FileOps.java
jperedadnr: Ok, done

src/test/java/com/gluonhq/substrate/util/FileOpsTests.java
jperedadnr: That's @johanvos doing :)

johanvos: It's a good test on support for i18n :)

src/main/java/com/gluonhq/substrate/util/FileOps.java
eugener: I meant `zis.closeEntry()` - it is a separate create/close cycle within the overall zip processing.

jperedadnr: I've realized I wasn't closing each entry. I've moved that to the `while`. The external try with resources will manage to call `zis.close()`.
I still don't see the need to include a try-catch for `zis.closeEntry()`.

eugener: I might be wrong, but `zis.closeEntry` is still perceptible to failure. Cannot say for sure what the effect of not closing the entry is - might be a memory leak. If you feel good about it - no need to fix

Issue #179, opened 2019-12-11
Add Initialize at build time flag


jperedadnr:
It can be used internally by Substrate for now, allowing build time initialization for instance of enums with large number of values

PR #182, opened 2019-12-11
#179 Add initialize at build time flag

jperedadnr:


src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -307,16 +308,16 @@ public boolean isUseLLVM() {
...
-    public List<String> getDelayInitList() {
-        return delayInitList;
+    public List<String> getInitBuildTimeList() {
+        return initBuildTimeList;
     }
 
     /**
      * Sets additional lists
-     * @param delayInitList a list of classes that will be added to the default delayed list
+     * @param initBuildTimeList a list of classes that will be added to the default delayed list
```
johanvos: that will be added to the initialize-at-buildtime list (as opposed to delayed)

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -106,6 +106,10 @@ public boolean compile(String cp) throws IOException, InterruptedException {
         useGlisten = GlistenResolver.useGlisten(cp);
         String nativeImage = getNativeImagePath();
         ProcessBuilder compileBuilder = new ProcessBuilder(nativeImage);
+        List<String> buildTimeList = getInitializeAtBuildTimeList(useGlisten);
```
johanvos: This is one for @eugener : do we want to have this boolean inside the method, or on the configuration?

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
jperedadnr: Fixed

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: Agree, we'll need a better way to handle the different "uses", but I'd say we can do that in a follow up PR?

eugener: Agree with @jperedadnr... we should try to redesign the resolvers using "strategy" pattern. But that can be done in a later PR - no sense of holding this one.

src/main/java/com/gluonhq/substrate/util/Strings.java
```diff
@@ -73,4 +76,12 @@ public static String substitute(String template, Map<String,String> context) {
         return substitute(template, context::get);
     }
 
+    /**
+     * Converts a string of comma separated strings into an array of strings
+     * @param s a string that contains comma separated strings
+     * @return a list of strings
+     */
+    public static List<String> splitString(String s) {
```
eugener: Good decision! I totally forgot about this one while creating the Strings class.

PR #184, opened 2019-12-11
#180 Add Attach classes to JNI file

jperedadnr:


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -636,6 +650,11 @@ String getAppPath(String appName) {
         return Collections.emptyList();
     }
 
+    // by default, false, but subclasses can decide against it.
```
johanvos: What is this method supposed to do/return?

jperedadnr: I've updated the javadoc

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @param archOs a string with the arch and os
+     * @return a list of classes that should be initialized at build time
+     * @throws IOException
+     */
+    public List<String> getUserInitBuildTimeList(String archOs) throws IOException {
+        Logger.logDebug("Scanning for init build time files");
+        return scanJars(USER_INIT_BUILD_TIME_FILE,
+                Strings.substitute(USER_INIT_BUILD_TIME_ARCHOS_FILE, Map.of("archOs", archOs)),
+                null,
```
johanvos: looks like we can use a Strings.substitute method with less parameters?

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -0,0 +1,149 @@
...
+     * META-INF/substrate/config/initbuildtime-${archos} files.
+     *
+     * The method will return a list of class names from all the files found
+     *
+     * @param archOs a string with the arch and os
+     * @return a list of classes that should be initialized at build time
+     * @throws IOException
+     */
+    public List<String> getUserInitBuildTimeList(String archOs) throws IOException {
+        Logger.logDebug("Scanning for init build time files");
```
eugener: Would be nice to check the `archOS` for null. Previous method does not need this as `Classpath` does it already.

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
eugener: @johanvos Not sure what you mean. I don't think we have a shorter one.

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -0,0 +1,149 @@
...
+     * META-INF/substrate/config/reflectionconfig-${archos} files.
+     *
+     * The method will return a list of lines from all the files found
+     *
+     * @param archOs a string with the arch and os
+     * @return a list of lines that should be added to the reflectionconfig.json file
+     * @throws IOException
+     */
+    public List<String> getUserReflectionList(String archOs) throws IOException {
+        Logger.logDebug("Scanning for reflection files");
```
eugener: Check the argument for null

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -0,0 +1,149 @@
...
+     * META-INF/substrate/config/jniconfig-${archos} files.
+     *
+     * The method will return a list of lines from all the files found
+     *
+     * @param archOs a string with the arch and os
+     * @return a list of lines that should be added to the jniconfig.json file
+     * @throws IOException
+     */
+    public List<String> getUserJNIList(String archOs) throws IOException {
+        Logger.logDebug("Scanning for JNI files");
```
eugener: Check the argument for null

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -0,0 +1,149 @@
...
+ * Helper class that helps scanning jars in the classpath looking for
+ * files in META-INF/substrate/config that contain information that has
+ * to be added to the config files and native image command line flags
+ */
+public class ConfigResolver {
+
+    private final List<File> jars;
+
+    public ConfigResolver(String classpath) {
+        this.jars = new ClassPath(classpath)
```
eugener: Since we require class path as a constructor parameter, why not to use `ClassPath`? 
I suspect that that is what the caller would have anyway?

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -0,0 +1,149 @@
...
+    public List<String> getUserJNIList(String archOs) throws IOException {
+        Logger.logDebug("Scanning for JNI files");
+        return scanJars(USER_JNI_FILE,
+                Strings.substitute(USER_JNI_ARCHOS_FILE, Map.of("archOs", archOs)),
+                ",",
+                line -> !line.startsWith("[") && !line.startsWith("]"));
+    }
+
+    private List<String> scanJars(String configName, String configArchosName, String initLine, Predicate<String> filter) throws IOException {
+        List<String> list = new ArrayList<>();
```
eugener: Check arguments for nulls

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
jperedadnr: Right, but this probably leads to a bigger refactoring. We have `processClassPath`, that returns a String, while it probably could return a `ClassPath` object, and that will be the only object we would need to create:
```
public boolean compile(String cp) throws IOException, InterruptedException {
        ClassPath classPath = processClassPath(cp);
    ...
```

Does that make sense? Probably for a follow-up PR then.

jperedadnr:
`ConfigResolver` depends on the classpath and `scanJars` will process all found jars in it. 
In order to provide some tests, I see two options: 
- add a dummy jar with some dummy content in `META-INF/substrate/config` to resources, then before the test, copy this jar in some tmp folder, and pass that path as "classpath", so we can process it and check the results.
- test-project/HelloGluon will use it in a real case scenario, once GM and Attach are released including the `META-INF/substrate/config` real content.

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
johanvos: then why don't we add a Strings.substitute() method with less parameters? If it is legal to call foo(A, B) with B null, I think it makes sense to add foo(A) that will default to foo(A, null). That looks cleaner to me than the caller passing null as an argument.

jperedadnr:
I've added unit tests, based on a dummy jar I've created and added to test/resources. 
@eugener Do you think it is necessary to keep the sources of that jar and how to generate it? It could be automated, but I don't think it is worth the hassle?

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
eugener: I think `null` is passed into `scanJars`, not into `substitute`?

eugener:
I don’t think source of the test jar is that important as it will hopefully never change.

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
johanvos: you're right. So that means my remark is about scanJars :)

eugener:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- src/main/java/com/gluonhq/substrate/config/ConfigResolver.java  8
- src/test/java/com/gluonhq/substrate/config/ConfigTests.java  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/gluonhq/substrate/pullRequest?prid=4678876&bid=15652398)

eugener:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- src/test/java/com/gluonhq/substrate/config/ConfigTests.java  1
- src/main/java/com/gluonhq/substrate/config/ConfigResolver.java  8
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/gluonhq/substrate/pullRequest?prid=4678876&bid=15652398)

PR #191, opened 2019-12-13
Added compilerArgs configuration option

lazar-mitrovic:
Added new configuration option to enable passing arbitrary arguments to native-image ( ` --verbose ` or ` --language:ruby ` for example)

### DIFF
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -308,6 +308,11 @@ public class InternalProjectConfiguration {
                 .orElse(Collections.emptyList());
     }
 
+    public List<String> getCompilerArgs() {
+        return Optional.ofNullable(publicConfig.getCompilerArgs())
+                .orElse(Collections.emptyList());
+    }
+
     public List<String> getInitBuildTimeList() {
         return initBuildTimeList;
     }
@@ -353,10 +358,7 @@ public class InternalProjectConfiguration {
     public String getMainClassName() {
         return publicConfig.getMainClassName();
     }
-
-    public String getCompilerArgs() {
-        return publicConfig.getCompilerArgs();
-    }
+    
 
     public IosSigningConfiguration getIosSigningConfiguration() {
         return Optional.ofNullable(publicConfig.getIosSigningConfiguration()).orElse(new IosSigningConfiguration());
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -354,11 +354,9 @@ public class InternalProjectConfiguration {
         return Objects.requireNonNull(publicConfig.getAppName(), "App name is required");
     }
 
-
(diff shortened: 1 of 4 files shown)


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -133,6 +135,8 @@ public boolean compile(String cp) throws IOException, InterruptedException {
         compileBuilder.command().add("-Dsvm.platform=org.graalvm.nativeimage.Platform$"+jniPlatform);
         compileBuilder.command().add("-cp");
         compileBuilder.command().add(classPath);
+        if (compilerArgs != null)
+            compileBuilder.command().add(compilerArgs);
```
jperedadnr: `compilerArgs` is a string, so only one argument will be allowed.
If more arguments are required, usually those can be added with a comma separated string. However I believe that won't work for the process builder. 
There is a convenient `Strings::split` method that could be used in this case, so the user's options will be added as a list.

lazar-mitrovic: My idea was that user could add space separated arguments (since ProcessBuilder would  eventually join all arguments into a single string anyway - through [ProcessImpl](https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/windows/classes/java/lang/ProcessImpl.java#L193))

However that on retrospect really seems hacky so I changed compilerArgs to be a list of strings in later commits.

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -353,6 +358,7 @@ public String getAppName() {
     public String getMainClassName() {
         return publicConfig.getMainClassName();
     }
+
```
jperedadnr: Remove blank line

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -96,6 +96,8 @@ public boolean compile(String cp) throws IOException, InterruptedException {
         FileOps.rmdir(paths.getTmpPath());
         String tmpDir = paths.getTmpPath().toFile().getAbsolutePath();
         String mainClassName = projectConfiguration.getMainClassName();
+        List<String> compilerArgs = projectConfiguration.getCompilerArgs();
+
```
jperedadnr: This is not necessary, add directly the compiler args to the process builder below

jperedadnr:
@lazar-mitrovic Thanks for your PR, btw.

According to our [Contribution guideliness](https://github.com/gluonhq/substrate#issues-and-contributions), we will need you to sign the [Contributor License Agreement](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY) before we can merge it.

lazar-mitrovic:
I filled out CLA form earlier today, but I did it again just in case I missed something.

jperedadnr:
Ok, thanks, we are still in the process of automating the CLA recognition.

PR #192, opened 2019-12-13
link to linux requirements when gcc/ld is too old

tiainen:


src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
```diff
@@ -133,16 +133,20 @@ private void checkLinker() throws InterruptedException, IOException {
     private void validateVersion(String[] processCommand, String processName, Version minimalVersion) throws InterruptedException, IOException {
         String versionLine = getFirstLineFromProcess(processCommand);
         if (versionLine == null) {
-            System.err.println("WARNING: we were unable to parse the version of your " + processName + ".\n" +
+            System.err.println(
```
eugener: In all cases below checks for `null` can be replaced with `Objects.requireNonNull(T obj, String message)`

tiainen: The version is allowed to be `null`, which is the case when the version could not be parsed from running i.e. `gcc --version`. In that case we try to use with the found compiler or linker. When using `Objects.requireNonNull()`, a `null` value will cause a `NullPointerException` to be thrown, which is not what we want here?

johanvos: True. In this case, we print a warning. In case the unknown compiler fails, the developer at least has this warning that gives a hint on what might have gone wrong.
In the long term, we should be able to detect the version in all cases. But that means we need to collect data from all corner cases, which will take lots of time.

PR #196, opened 2019-12-13
define a single regex to include resources by extension

tiainen:


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -439,22 +439,19 @@ Path getLlcPath() throws IOException {
...
             "license", "json");
 
-    private  List<String> getResources() {
-        List<String> resources = new ArrayList<>(resourcesList);
+    private List<String> getResources() {
+        List<String> resources = RESOURCES_BY_EXTENSION.stream()
+                .map(extension -> "-H:IncludeResources=.*\\." + extension + "$")
+                .collect(Collectors.toList());
+
         resources.addAll(projectConfiguration.getResourcesList());
```
jperedadnr: What happens with the user's list? it won't have `-H:IncludeResources=`. The user should only provide, when needed, a list with file extensions.

tiainen: Good catch, I've added the prefix back in.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -439,22 +439,24 @@ Path getLlcPath() throws IOException {
...
-        resources.addAll(projectConfiguration.getResourcesList());
+    private List<String> getIncludeResourcesArguments() {
+        List<String> resourcesByExtension = RESOURCES_BY_EXTENSION.stream()
+                .map(extension -> "-H:IncludeResources=.*\\." + extension + "$")
+                .collect(Collectors.toList());
 
-        List<String> list = resources.stream()
-                .map(s -> "-H:IncludeResources=.*/.*" + s + "$")
+        List<String> configurationResources = projectConfiguration.getResourcesList().stream()
+                .map(resource -> "-H:IncludeResources=" + resource)
```
jperedadnr: This implies the user will need to provide a regex too? I'm okay with this, as it will be more flexible probably, but we'll need to update the docs about it.

tiainen: It's indeed more flexible, and the docs currently say the following:

> List of additional resource patterns or extensions that will be added...

It might be an improvement to explicitly mention that it should be a regular expression pattern.

jperedadnr: We should do a follow-up on this, even adding a small example to the docs.

PR #228, opened 2019-12-27
Remove redundant checks in `SubstrateDisdpatcher`

eugener:


src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -179,28 +180,19 @@ private TargetConfiguration getTargetConfiguration(Triplet targetTriplet) {
      */
     public boolean nativeCompile(String classPath) throws Exception {
         config.canRunNativeImage();
-        if (classPath != null) {
```
eugener: No need to check for `null` as it happens within `ClassPath` constructor

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -179,28 +180,19 @@ private TargetConfiguration getTargetConfiguration(Triplet targetTriplet) {
...
-            config.setUseJavaFX(useJavaFX);
-        }
+        boolean useJavaFX = new ClassPath(classPath).contains( s -> s.contains("javafx"));
+        config.setUseJavaFX(useJavaFX);
 
         Triplet targetTriplet  = config.getTargetTriplet();
         if (!config.getHostTriplet().canCompileTo(targetTriplet)) {
             throw new IllegalArgumentException("We currently can't compile to "+targetTriplet+" when running on "+config.getHostTriplet());
         }
-        if (targetConfiguration == null) {
```
eugener: Same here as the null check happens within Dispatcher's constructor

Issue #232, opened 2019-12-30
Allow ConfigResolver scan classes folder


jperedadnr:
So far we only scan jars, but if the project contains config files, those will be in target/classes.

Creating a jar out of that folder is an easy way.

PR #233, opened 2019-12-30
#232 Create a jar from classes folder that can be scanned

jperedadnr:
for user’s configuration files

### DIFF
--- a/src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
+++ b/src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
@@ -30,10 +30,13 @@ package com.gluonhq.substrate.config;
 import com.gluonhq.substrate.model.ClassPath;
 import com.gluonhq.substrate.util.FileOps;
 import com.gluonhq.substrate.util.Logger;
+import com.gluonhq.substrate.util.ProcessRunner;
 import com.gluonhq.substrate.util.Strings;
 
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -114,6 +114,7 @@ public boolean compile(String cp) throws IOException, InterruptedException {
         }
         Path gvmPath = paths.getGvmPath();
         FileOps.rmdir(paths.getTmpPath());
+        Files.createDirectories(paths.getTmpPath());
```
johanvos: Don't we have generic code that makes sure the required paths are created?

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
```diff
@@ -61,11 +64,35 @@
 
     private final List<File> jars;
 
-    public ConfigResolver(String classpath) {
-        this.jars = new ClassPath(classpath)
+    /**
+     * ConfigResolver constructor
+     *
+     * @param classpath a string with the full classpath of the user's project
+     * @param tmpPath a temporal path where classes.jar will be added
```
johanvos: is tmpPath needed outside of this method? If not, it seems better to create/delete it inside this method?

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: We do, but the line before just removes that path

src/main/java/com/gluonhq/substrate/config/ConfigResolver.java
jperedadnr: We just need a temporal path, and using `client/archOS/gvm/tmp` seems convenient, but then we need to pass that path.

johanvos: Don't we have Paths.getTmpPath or so?

jperedadnr: No, we don't. Well, we could use Files.createTempDirectory, outside client/archOS/gvm

johanvos: If we don't need the results outside of this method, Files.createTempDirectory sounds better.

Issue #165, opened 2019-12-06
iOS-deploy not flushing immediately the output


jperedadnr:
When running `mvn client:run` on iOS, iOS-deploy should flush the output of the application to the console, but it seems to keep it in a buffer and flush it from time to time (more than 10 seconds).

In the meantime, this will work from a terminal:

`/usr/local/bin/ios-deploy  --bundle <path/to>/target/client/arm64-ios/<appname>.app --no-wifi --debug --noninteractive`

PR #236, opened 2020-01-03
#165 Use local formula to install from Gluon’s fork of ios-deploy

jperedadnr:
This PR allows verbose debug from mvn client:run.

Caveats: We need to maintain our [fork](https://github.com/gluonhq/ios-deploy.git) in sync with upstream
We need to maintain the local formula from [upstream](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ios-deploy.rb) (it is not that critical as we will be installing always from head)

Also, when checking for iOS-deploy, if it is already installed, the new local version won't be installed, this requires user intervention to uninstall iOS-deploy first.


johanvos:
> Caveats: We need to maintain our [fork](https://github.com/gluonhq/ios-deploy.git) in sync with upstream

This is not a caveat but a strong requirement, I believe.

PR #249, opened 2020-01-07
Use latest Android build-tools

### LINKED ISSUE #255: AndroidTargetConfiguration has hard-coded values for Build Tools
AndroidTargetConfiguration uses hard-coded version i.e. `27.0.3` for build-tools. This forces Substrate to work only if the said version of the build-tools is installed. This is highly inflexible. Substrate should be able to use any version of build-tools present. 

To minimize unexpected behaviour, using the latest among all present build-tools versions is recommended.

abhinayagarwal:
Removes hard-coded version of build-tools and uses the latest version present in Android SDK path.

Fixes #255

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, Gluon
+ * Copyright (c) 2019, 2020, Gluon
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -29,10 +29,11 @@ package com.gluonhq.substrate.target;
 
 import com.gluonhq.substrate.Constants;
 import com.gluonhq.substrate.model.ClassPath;
-import com.gluonhq.substrate.model.ProcessPaths;
 import com.gluonhq.substrate.model.InternalProjectConfiguration;
+import com.gluonhq.substrate.model.ProcessPaths;
 import com.gluonhq.substrate.util.FileOps;
 import com.gluonhq.substrate.util.ProcessRunner;
+import com.gluonhq.substrate.util.Version;
 
 import java.io.IOException;
 import java.nio.file.Files;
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -360,4 +362,20 @@ private void createDevelopKeystore() throws IOException, InterruptedException {
 
         System.err.println("done creating ks");
     }
+    
+    private String findLatestBuildTool(Path sdkPath) throws IOException {
+        Path buildToolsPath = sdkPath.resolve("build-tools");
+        if (buildToolsPath.toFile().exists()) {
```
jperedadnr: Use `Files.exists(buildToolsPath)` instead

src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -360,4 +362,20 @@ private void createDevelopKeystore() throws IOException, InterruptedException {
 
         System.err.println("done creating ks");
     }
+    
+    private String findLatestBuildTool(Path sdkPath) throws IOException {
+        Path buildToolsPath = sdkPath.resolve("build-tools");
```
jperedadnr: `Objects.requireNonNull(sdkPath)` and `Files.exists(sdkPath)`?

src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -360,4 +362,20 @@ private void createDevelopKeystore() throws IOException, InterruptedException {
...
+        if (buildToolsPath.toFile().exists()) {
+            Optional<Version> latest = Files.walk(buildToolsPath, 1)
+                    .filter(file -> Files.isDirectory(file) && !file.equals(buildToolsPath))
+                    .map(file -> new Version(file.getFileName().toString()))
+                    .max(Version::compareTo);
+            if (latest.isPresent()) {
```
jperedadnr: maybe you simplify it to something like `return Files.walk().filter().map().max(...).map(toString).orElseThrow(IOException)`?

src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
abhinayagarwal: Adding a check for `Files.exists(buildToolsPath)` makes `Files.exists(sdkPath)` redundant.

PR #281, opened 2020-01-17
#280 Copy other assets from the iOS folder

### LINKED ISSUE #280: All items of the iOS/assets folders should be included
Currently only the assets from the root `ios/assets` and the `ios/assets/Assets.xcassets` folder are included.

jperedadnr:
Fixes #280

### DIFF
--- a/src/main/java/com/gluonhq/substrate/util/ios/InfoPlist.java
+++ b/src/main/java/com/gluonhq/substrate/util/ios/InfoPlist.java
@@ -363,6 +363,13 @@ public class InfoPlist {
         }
     }
 
+    /**
+     * Scans the ios/assets folder for possible folders other than
+     * Assets.cassets (which is compressed with actool),
+     * and copy them directly to the app folder
+     * @param resourcePath the path for ios assets
+     * @throws IOException
+     */
     private void copyOtherAssets(Path resourcePath) throws IOException {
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/util/ios/InfoPlist.java
```diff
@@ -361,4 +363,22 @@ private void verifyAssetCatalog(Path resourcePath, String platform, String minOS
         }
     }
 
+    private void copyOtherAssets(Path resourcePath) throws IOException {
```
johanvos: add some doc: this is copying $project/src/main/resources/assets/foo to $project/build bar?

src/main/java/com/gluonhq/substrate/util/ios/InfoPlist.java
```diff
@@ -363,6 +363,13 @@ private void verifyAssetCatalog(Path resourcePath, String platform, String minOS
         }
     }
 
+    /**
+     * Scans the ios/assets folder for possible folders other than
```
johanvos: from this, it is not clear to me if that assets folder is something the user has to supply, or that it is auto-generated.
If the user has to supply it, it should start with src
if it is generated, it should start with gensrc

PR #297, opened 2020-01-27
Supporting custom native code in project

lazar-mitrovic:
This PR adds support for `native` directory in project which can contain native `.h` or `.c` files which would be included for compilation and linking. Without this modification you either need to insert files in `target` manually or to rebuild Substrate each time you make changes in native code. 

Example usage can be seen here: https://github.com/lazar-mitrovic/HelloNative

PR also includes `jni.h` location for linux target.

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
@@ -769,11 +769,11 @@ public abstract class AbstractTargetConfiguration implements TargetConfiguration
         return Arrays.asList("-o", getAppPath(appName));
     }
 
-    Path getNativeCodePath() {
+    protected Path getNativeCodePath() {
         return paths.getSourcePath().getParent().resolve("native");
     }
 
-    List<String> getNativeCodeList() throws IOException {
+    protected List<String> getNativeCodeList() throws IOException {
         Path nativeCodeDir = getNativeCodePath();
         if (!Files.exists(nativeCodeDir))
             return Collections.emptyList();
--- a/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
@@ -773,7 +773,7 @@ public abstract class AbstractTargetConfiguration implements TargetConfiguration
         return Arrays.asList("-o", getAppPath(appName));
     }
 
-    List<String> getTargetNativeCodeExtensions() {
+    protected List<String> getTargetNativeCodeExtensions() {
         return Arrays.asList(".c");
     }
 
--- a/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
@@ -41,7 +41,6 @@ import com.gluonhq.substrate.util.Strings;
 
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
-import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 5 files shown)


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -756,6 +769,20 @@ String getAppPath(String appName) {
         return Arrays.asList("-o", getAppPath(appName));
     }
 
+    protected Path getNativeCodePath() {
```
johanvos: this can be moved to com.gluonhq.substrate.model.ProcessPaths

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -317,11 +323,18 @@ protected boolean compileAdditionalSources()
             FileOps.copyResource(getAdditionalSourceFileLocation()  + fileName, workDir.resolve(fileName));
             processBuilder.command().add(fileName);
         }
+        
+        Path nativeCodeDir = getNativeCodePath();
+        for( String fileName: getNativeCodeList() ) {
+            FileOps.copyFile(nativeCodeDir.resolve(fileName), workDir.resolve(fileName));
+            processBuilder.command().add(fileName);
```
johanvos: this would also bring the header files to the compile command. We need to copy the header files, but they should not be in the compile command line.
e.g. we need `gcc -I/path/to/headers hello.c` 
with `hello.c` having `#include hello.h` and hello.h needs to be at path/to/headers

I just found out that the current code we have for this is wrong (it uses the header file in the gcc argument, which is not what we want)

lazar-mitrovic: Ah I see. I was relying on current code for reference. 
This should be fixed in latest commit.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -313,15 +319,26 @@ protected boolean compileAdditionalSources()
...
+
+        processBuilder.command().add("-I" + workDir.toString());
+
         for( String fileName: getAdditionalSourceFiles() ) {
             FileOps.copyResource(getAdditionalSourceFileLocation()  + fileName, workDir.resolve(fileName));
             processBuilder.command().add(fileName);
         }
+        
+        Path nativeCodeDir = paths.getNativeCodePath();
+        if (Files.isDirectory(nativeCodeDir))
```
jperedadnr: I'd rather use curly braces on if statements

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -756,6 +773,16 @@ String getAppPath(String appName) {
         return Arrays.asList("-o", getAppPath(appName));
     }
 
+    protected List<String> getNativeCodeList() throws IOException {
+        Path nativeCodeDir = paths.getNativeCodePath();
+        if (!Files.exists(nativeCodeDir))
```
jperedadnr: same here (curly braces)

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -756,6 +773,16 @@ String getAppPath(String appName) {
...
+            return Collections.emptyList();
+        return Files.list(nativeCodeDir)
+            .map(p -> p.getFileName().toString())
```
jperedadnr: check non null objects, non null file name, if that make sense?

lazar-mitrovic: `paths.getNativeCodePath()` always returns non null value (since it only appends `native` on project path)
Directory is checked on line 778, so `Files.list(nativeCodeDir)` will be either `Stream` or empty `Stream`.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -756,6 +773,16 @@ String getAppPath(String appName) {
...
+            .filter(s -> s.endsWith(".c"))
```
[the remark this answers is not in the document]
lazar-mitrovic: This should be implemented in latest commit

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
jperedadnr: ok

PR #301, opened 2020-01-29
Various android improvements

lazar-mitrovic:
This PR brings makes Android target behave more like iOS target:
- `runUntilEnd` installs application, **and then** starts it and shows log
- apps that don't use JFX (`stdout` only apps) now don't crash in runtime

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
@@ -253,7 +253,6 @@ public class AndroidTargetConfiguration extends PosixTargetConfiguration {
                 ProcessRunner log = new ProcessRunner(sdkPath.resolve("platform-tools").resolve("adb").toString(),
                 "-d", "logcat", "-v", "brief", "-v", "color", "GraalCompiled:V", "GraalGluon:V", "AndroidRuntime:E", "ActivityManager:W", "*:S");
                 log.setInfo(true);
-                System.out.println(log.getCmd());
                 log.runProcess("log");
             } catch (IOException | InterruptedException e) { 
                 e.printStackTrace(); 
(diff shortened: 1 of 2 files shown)


src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -242,7 +242,32 @@ public boolean runUntilEnd() throws IOException, InterruptedException {
...
+        if (processResult != 0) throw new IOException("Application instalation failed!");
+
+        Runnable logcat = () -> {
+            try {
+                ProcessRunner clearLog = new ProcessRunner(sdkPath.resolve("platform-tools").resolve("adb").toString(),
+                "logcat", "-c");
+                clearLog.runProcess("clearLog");
+
+                ProcessRunner log = new ProcessRunner(sdkPath.resolve("platform-tools").resolve("adb").toString(),
+                "-d", "logcat", "-v", "brief", "-v", "color", "GraalCompiled:V", "GraalGluon:V", "AndroidRuntime:E", "ActivityManager:W", "*:S");
```
johanvos: this is a nice feature, but I wonder if we should have an option to toggle it on/off (the whole logcat thread). For example, I typically use logcat -v threadtime but depending on what you are doing, you want to try different options.

lazar-mitrovic: `threadtime` would display the date and invocation time which are already added by ProcessRunner. I was aiming for most iOS-like option without losing vital information (since Android has no per app filter, several different tags needed to be included in order to catch crash logs). [Here](https://stackoverflow.com/a/48479579) is the nice breakdown of `logcat` format options, so we can decide whether there is more suitable option.

About toggling whole thread I guess we could use `verbose` configuration option, however I didn't see it used for runtime options yet, and `CTRL+C` would kill the `logcat` thread anyway.

johanvos: I was actually referring to make the whole log process toggled on/off.
As for the treadtime option: thread shows the process + parent process on android, and ProcessRunner doesn't have that. The time from the ProcessRunner is the received time, which can be different from the "android time".

Combining these remarks: the options here are perfect, and they fit well with the ios style. 

Thinking about it for a bit longer: you're totally right this should not even be an option. Android is the only platform where "client:run" didn't result in a run, so it has to be changed to be in line with the others. We are thinking about other lifecycle events (e.g. install, package), but that is another issue.

jperedadnr: I guess this option will be the default one you get with `client:run`, but you can always go to a terminal and run `adb` with other options if needed.

PR #304, opened 2020-01-30
Increase number of retries and delay b/w each retry

abhinayagarwal:
Last 2 releases have failed because the nexus-staging plugin wasn't able to close and release the repository.

Increasing number of retries and wait time b/w each retry might be useful to tackle this situation.


build.gradle
```diff
@@ -113,4 +113,6 @@ nexusPublishing {
 nexusStaging {
     username = project.hasProperty('sonatypeUsername') ? project.property('sonatypeUsername') : ''
     password = project.hasProperty('sonatypePassword') ? project.property('sonatypePassword') : ''
+    numberOfRetries = 30
+    delayBetweenRetriesInMillis = 3000
```
johanvos: isn't this a bit short? In case there is an issue at sonatype, it typically last for a longer time.

PR #358, opened 2020-02-18
 #357 Bump target SDK version to 28

### LINKED ISSUE #357: AndroidManifest target is too low
Android target SDK version is set in the default AndroidManifest to 21, and that triggers a warning when installing the app.

Target has to be set at least to 26 (28 since latest requirements to submit to Play Store)

jperedadnr:
Also Rename namespace from `xmlns:a` to `xmlns:android`

Fixes #357

### DIFF
--- a/src/main/resources/native/android/AndroidManifest.xml
+++ b/src/main/resources/native/android/AndroidManifest.xml
@@ -1,12 +1,12 @@
 <?xml version='1.0'?>
-<manifest xmlns:android='http://schemas.android.com/apk/res/android' package='com.gluonhq.helloandroid' android:versionCode='1' android:versionName='1.0'>
-    <supports-screens android:xlargeScreens="true"/>
-    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="28"/>
-    <application android:label='A HelloGraal' android:icon="@mipmap/ic_launcher">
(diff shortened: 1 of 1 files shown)


johanvos:
Did you check that this work?
I am afraid that as of 28, a library can not have R+X segments, which our generated lib does.

jperedadnr:
I've checked HelloFX on my Pixel X with Android 10. 
Note that this is now mandatory for publishing apps to the store.

johanvos:
Let's wait for the next GraalVM build, to see if the R/X segments are separated.

src/main/resources/native/android/AndroidManifest.xml
```diff
@@ -1,12 +1,12 @@
 <?xml version='1.0'?>
-<manifest xmlns:a='http://schemas.android.com/apk/res/android' package='com.gluonhq.helloandroid' a:versionCode='1' a:versionName='1.0'>
```
johanvos: I don't think renaming a to android adds anything, so can this be reverted?

jperedadnr: sure

jperedadnr: done

Issue #403, opened 2020-03-04
remove CustomLLC flag


johanvos:
Since GraalVM is now bundled with llvm, including llc, there is no need to supply a customLLC flag.

tiainen:
Fixed with #410

PR #410, opened 2020-03-06
#403 remove custom llc

tiainen:
Make use of LLVM that is shipped with GraalVM instead of downloading a custom version.

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
@@ -297,7 +297,6 @@ public class AndroidTargetConfiguration extends PosixTargetConfiguration {
     @Override
     List<String> getTargetSpecificAOTCompileFlags() throws IOException {
         return Arrays.asList("-H:CompilerBackend=" + Constants.BACKEND_LLVM,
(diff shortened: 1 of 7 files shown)


src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -296,18 +296,14 @@ boolean allowHttps() {
 
     @Override
     List<String> getTargetSpecificAOTCompileFlags() throws IOException {
-        Path llcPath = getLlcPath();
-        Path internalLlcPath = projectConfiguration.getGraalPath().resolve("lib").resolve("llvm").resolve("bin");
-        
         return Arrays.asList("-H:CompilerBackend=" + Constants.BACKEND_LLVM,
+                "--features=org.graalvm.home.HomeFinderFeature",
```
johanvos: this should go to AbstractTargetConfiguration

tiainen: Should we enable the feature unconditionally or only when LLVM backend is used?

johanvos: unconditionally. At this moment, it is by default enabled in GraalVM for the LIR backend, but defaults can change.

PR #413, opened 2020-03-07
Notify on openURL calls for iOS

stevesoltys:
This provides Attach services the ability to listen for `openURL` calls and handle them accordingly.

An example use case for this is Firebase, where the module needs to handle this callback from Google Sign In: https://developers.google.com/identity/sign-in/ios/sign-in?ver=objc#enable_sign-in

I'm not sure if this is the best approach for this, let me know if there's something better that I could do in the Attach service itself :)

### DIFF
--- a/src/main/resources/native/ios/AppDelegate.m
+++ b/src/main/resources/native/ios/AppDelegate.m
@@ -109,6 +109,32 @@ int main(int argc, char * argv[]) {
     gvmlog(@"[UIAPP] applicationWillTerminate");
 }
 
+- (BOOL)application:(UIApplication *)app
+            openURL:(NSURL *)url
+            options:(NSDictionary<NSString *, id> *)options {
+
+    gvmlog(@"[UIAPP] application_openURL_options");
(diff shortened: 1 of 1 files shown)


src/main/resources/native/ios/AppDelegate.m
```diff
@@ -109,6 +109,32 @@ - (void)applicationWillTerminate:(UIApplication *)application {
...
+            openURL:(NSURL *)url
+            options:(NSDictionary<NSString *, id> *)options {
+
+    gvmlog(@"[UIAPP] application_openURL_options");
+
+    [[NSNotificationCenter defaultCenter]
+        postNotificationName:@"UIApplicationOpenURLNotification"
+        object:self userInfo:[NSDictionary dictionaryWithObject:url forKey:@"URL"]];
+
+    return YES;
```
stevesoltys: Not sure what the best approach would be for knowing whether or not the URL notification was actually handled.. or if it matters.

PR #431, opened 2020-03-11
Define git identity for Gluon Bot

abhinayagarwal:
The last GA deploy's post-release step failed complaining that [Git identity](https://github.com/gluonhq/substrate/runs/500486973?check_suite_focus=true#step:7:12) was not set. This PR fixes the same.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -44,8 +44,10 @@ jobs:
       - name: Commit next development version
         if: steps.deploy.outputs.exit_code == 0
         run: |
+          git config user.email "githubbot@gluonhq.com"
+          git config user.name "Gluon Bot"
           TAG=${GITHUB_REF/refs\/tags\//}
           newVersion=${TAG%.*}.$((${TAG##*.} + 1)) # Update version by 1
           sed -i -z "0,/version = $TAG/s//version = $newVersion-SNAPSHOT/" gradle.properties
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -44,8 +44,10 @@ jobs:
...
         if: steps.deploy.outputs.exit_code == 0
         run: |
+          git config user.email "githubbot@gluonhq.com"
+          git config user.name "Gluon Bot"
           TAG=${GITHUB_REF/refs\/tags\//}
           newVersion=${TAG%.*}.$((${TAG##*.} + 1)) # Update version by 1
           sed -i -z "0,/version = $TAG/s//version = $newVersion-SNAPSHOT/" gradle.properties
-          git commit gradle.properties -m "Upgrade version to $newVersion-SNAPSHOT" --author "Gluon Bot <githubbot@gluonhq.com>"
+          git commit gradle.properties -m "Prepare development of $newVersion"
           git push https://gluon-bot:$GITHUB_PASSWORD@github.com/$GITHUB_REPOSITORY HEAD:master
```
tiainen: I am not sure, but we probably still lack required authentication to be able to push the changes. This might be an interesting thread should this PR still be insufficient: https://github.community/t5/GitHub-Actions/how-does-one-commit-from-an-action/td-p/30333

abhinayagarwal: Thanks for pointing it out. Will cross-verify and update if required.

abhinayagarwal: I have tested this and it works out well :)

PR #473, opened 2020-03-26
Fix for Watchdog issue

lazar-mitrovic:
* Fixes https://github.com/oracle/graal/issues/2283
* added `exception stack trace` support


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -85,7 +85,11 @@
             "--no-server",
             "-H:+ExitAfterRelocatableImageWrite",
             "-H:+SharedLibrary",
-            "-H:+AddAllCharsets"
+            "-H:+AddAllCharsets",
+            "-H:+ReportExceptionStackTraces",
+            "-H:+AllowIncompleteClasspath",
```
johanvos: Why do we need AllowIncompleteClasspath for this?

lazar-mitrovic: Removed now (as decided during slack conversation)

Issue #485, opened 2020-03-31
Refactor Lifecycle events


jperedadnr:
Related to https://github.com/gluonhq/attach/issues/86

PR #486, opened 2020-03-31
#485 Allow getting the Dalvik lifecycle class

jperedadnr:
No need to pass JNIEnv

### DIFF
--- a/src/main/resources/native/android/c/attach_adapter.c
+++ b/src/main/resources/native/android/c/attach_adapter.c
@@ -28,6 +28,7 @@
 #include "grandroid.h"
 
 jclass jUtilClass;
+jclass jLifecycleServiceClass;
 jclass jBleServiceClass;
 jclass jKeyboardServiceClass;
 jclass jPositionServiceClass;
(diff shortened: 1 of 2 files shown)


src/main/resources/native/android/c/attach_adapter.c
```diff
@@ -86,5 +91,5 @@ JNIEXPORT void JNICALL Java_com_gluonhq_helloandroid_MainActivity_nativeDispatch
 JNIEXPORT void JNICALL Java_com_gluonhq_helloandroid_MainActivity_nativeDispatchActivityResult(JNIEnv *env, jobject activity, jint requestCode, jint resultCode, jobject intent)
 {
     LOGE(stderr, "Dispatching activity result from native Dalvik layer: %d %d", requestCode, resultCode);
-    attach_setActivityResult(env, requestCode, resultCode, intent);
+    attach_setActivityResult(requestCode, resultCode, intent);
```
johanvos, line 94: I think this indicates we no longer need to method that has the JNIEnv?

PR #530, opened 2020-04-17
Fix for "Garbage-collected heap size exceeded." error on Android

lazar-mitrovic:


src/main/resources/native/android/c/launcher.c
```diff
@@ -59,7 +59,8 @@ const char *origargs[] = {
     "-Dcom.sun.javafx.gestures.scroll=true",
     "-Djavafx.verbose=true",
     "-Dmonocle.input.traceEvents.verbose=true",
-    "-Dprism.verbose=true"};
+    "-Dprism.verbose=true",
+    "-Xmx4g"};
```
jperedadnr, line 63: You need to increase the `argsize` below.

lazar-mitrovic: You are right, I fixed it in latest commit so that it calculates number of arguments automatically.

Issue #564, opened 2020-04-30
GitHub Actions installs libusbmuxd and libimobiledevice


tiainen:
Find out why we need to install the development version of libusbmux and libimobiledevice in GitHub Actions:

> brew install --HEAD libusbmuxd libimobiledevice

tiainen:
These libraries are only required when a developer wants to run the application on a real iOS device. Substrate makes use of the tool `ios-deploy` to deploy and debug the application on the device. And this tool depends on the libraries `usbmuxd` and `imobiledevice`.

In any case, the CI jobs don't deploy the application to a physical device, so there is no reason for them to install these libraries.

PR #565, opened 2020-04-30
Investigate reason for installing libusbmuxd and libimobiledevice

tiainen:
Find out why we need to install libusbmuxd and libimobiledevice.


tiainen:
There is no reason for these libraries during the CI jobs, so I've removed the install command.

tiainen:
Fixes #564

PR #567, opened 2020-04-30
pass CLI args to java main method

### LINKED ISSUE #566: CLI arguments aren't fed to java main
CLI arguments that are passed along with the native executable are not fed into the java main method.

tiainen:
This PR fixes #566 making sure to pass down all CLI arguments to the java main method.

### DIFF
--- a/src/main/resources/native/macosx/AppDelegate.m
+++ b/src/main/resources/native/macosx/AppDelegate.m
@@ -28,9 +28,7 @@
 #import <Cocoa/Cocoa.h>
 #include <pthread.h>
 
-extern void *run_main(int argc, char* argv[]);
-
-const char* args[] = {"myapp"};
+extern void *run_main(int argc, const char* argv[]);
 
 @interface AppDelegate : NSObject <NSApplicationDelegate>
 @end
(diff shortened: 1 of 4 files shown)


eugener:
Codacy Here is an overview of what got changed by this pull request:


```diff

Clones added
============
- src/main/resources/native/linux/launcher.c  1
- src/main/resources/native/windows/launcher.c  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/gluonhq/substrate/pullRequest?prid=5451961&bid=17874714)

src/main/resources/native/macosx/AppDelegate.m
```diff
@@ -28,9 +28,7 @@
 #import <Cocoa/Cocoa.h>
 #include <pthread.h>
 
-extern void *run_main(int argc, char* argv[]);
-
-const char* args[] = {"myapp"};
```
jperedadnr, line 33: We are removing `myapp`, which seems not needed at all. It is still present in the iOS and Android args, though. But maybe for another issue.

tiainen: While launching the application, when the main() method gets invoked, the first argument that is being passed is always the name of the executable that is being run. I don't think Graal does anything specific with that argument. What I have done in this PR is to basically replace `"myapp"` with `argv[0]`.

I have no idea how iOS and android are being launched by the OS. In any case, you can't launch the application from command line and pass in arbitrary command line arguments on those systems anyway, so the issue that is being fixed by this PR doesn't apply there.

PR #591, opened 2020-05-15
#588 Remove quarantine from root

### LINKED ISSUE #588: Notarize starting from root folder if path contains Contents/Home

jperedadnr:
Fixes #588.

If `GRAALVM_HOME` ends with `Contents/Home`, we remove quarantine starting from that path, but the parent folder remains with quarantine.

This PR removes quarantine from root GraalVM folder.

### DIFF
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -461,13 +461,18 @@ public class InternalProjectConfiguration {
         if (graalvmHome == null || graalvmHome.isEmpty()) {
             return;
         }
-        Logger.logDebug("Checking execution permissions for " + graalvmHome);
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -461,21 +461,26 @@ private void checkGraalVMPermissions(String graalvmHome) {
         if (graalvmHome == null || graalvmHome.isEmpty()) {
             return;
         }
-        Logger.logDebug("Checking execution permissions for " + graalvmHome);
+        String graalvmRoot = graalvmHome.endsWith("Contents/Home") ?
```
tiainen, line 464: Checking for `/Contents/Home` is a little bit more correct.

jperedadnr: Makes sense, fixed

PR #607, opened 2020-05-20
Use GraalVM 20.1 in Github Actions

abhinayagarwal:
Use latest release of GraalVM in Github Actions


jperedadnr:
If we do this change we won't be testing again latest builds that are created daily. Is this something that we want?

abhinayagarwal:
> 
> 
> If we do this change we won't be testing again latest builds that are created daily. Is this something that we want?

Good question. I will leave this on @johanvos to decide.

tiainen:
Regression is tested by our internal integration tests. They are run nightly against all supported platforms.

PR #629, opened 2020-05-27
Don’t require LLVM toolchain on Android with LIR backend

### LINKED ISSUE #625: Only install LLVM if needed
On Android, backend can be LIR, so that won't need LLVM toolchain.

jperedadnr:
Fixes #625

### DIFF
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -444,8 +444,8 @@ public class InternalProjectConfiguration {
             // host doesn't use LLVM
             return;
         }
-        if (new Triplet(Constants.Profile.ANDROID).equals(triplet) && getBackend().equals(Constants.BACKEND_LIR)) {
-            // Android with LIR doesn't use LLVM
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -441,6 +441,11 @@ public void canRunNativeImage() throws IOException, InterruptedException {
     public void canRunLLVM(Triplet triplet) throws IOException, InterruptedException {
         if (!new Triplet(Constants.Profile.IOS).equals(triplet) &&
                 !new Triplet(Constants.Profile.ANDROID).equals(triplet)) {
+            // host doesn't use LLVM
+            return;
+        }
+        if (new Triplet(Constants.Profile.ANDROID).equals(triplet) && getBackend().equals(Constants.BACKEND_LIR)) {
```
johanvos, line 447: Better to reverse that test (if backend is not LLVM, return).
There might be different non-LLVM backends, or at least different names for this backend.

jperedadnr: Done

PR #651, opened 2020-06-04
Wait for a minute before we release the client plugins

abhinayagarwal:
One minute of wait time will give Maven Central time to sync and make Substrate artifact available to be used by the plugins.

### DIFF
--- a/.github/scripts/release-plugins.sh
+++ b/.github/scripts/release-plugins.sh
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+echo "Wait for a minute for Maven Central to sync and make the artifact available to client plugins"
+counter=1
+while [ $counter -le 60 ]
+do
+  printf "."
+  sleep 1s
+((counter++))
+done
+
 git config --global user.email "githubbot@gluonhq.com"
 git config --global user.name "Gluon Bot"
 


.github/scripts/release-plugins.sh
```diff
@@ -1,5 +1,14 @@
 #!/usr/bin/env bash
 
+echo "Wait for a minute for Maven Central to sync and make the artifact available to client plugins"
+counter=1
+while [ $counter -le 60 ]
+do
+  printf "."
+  sleep 1s
+((counter++))
+done
```
jperedadnr, line 10: add a return carriage?

tiainen: There is absolutely no need to print progress, since this is run on a CI environment. Just print a message right before the sleep and then sleep 60s instead of using the counter.

abhinayagarwal: Even on a CI, it is weird to wait for 60s without any feedback. Instead of printing '.', we could also print the time elapsed.

tiainen: That's why I suggested to "print a message right before". Something like:
> Sleeping 60 seconds to allow Maven Central to do their thing.

On another note, why did you choose 60 seconds? AFAIK the sync can take up to 10 minutes. Perhaps it's time to just abandon trying to release both substrate and the client plugins from a single script.

abhinayagarwal: I presumed 60seconds should be good enough and it takes time for builds in Travis for client plugins to start as well. I restarted the failed builds immediately after they failed and they passed without any issue.

bq. Perhaps it's time to just abandon trying to release both substrate and the client plugins from a single script.

This change was [introduced recently](https://github.com/gluonhq/substrate/commit/152f2f3772871d64a263a009aaf41bf49dba4d0a) to ease the requirement of releasing Substrate and plugins together.

tiainen: Oops, now I see that I overlooked the echo command that you already added.

Anyway, if you believe these dots are useful, you can have them. I'm not the one that is going to look at live CI output :smile:.

PR #662, opened 2020-06-10
Extract and add aar files to the Android project

### LINKED ISSUE #660: Use aar libraries
Scan Attach jars for aar files, and add them to the android project.

jperedadnr:
Follow-up of https://github.com/gluonhq/attach/pull/164

Using aar libraries with the dalvik classes and their own manifest removes the need of manually adding permissions, as the final project manifest results from merging all existing manifests.

CopyOtherDalvikClasses is still in use, in case there are third party jars that bundle precompiled dalvik classes, but not in an aar file.

Fixes #660

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
@@ -125,6 +125,7 @@ public class AndroidTargetConfiguration extends PosixTargetConfiguration {
         prepareAndroidProject();
         prepareAndroidManifest();
         prepareAndroidResources();
+        copyAarLibraries();
         copyOtherDalvikClasses();
         copySubstrateLibraries();
         String configuration = generateSigningConfiguration();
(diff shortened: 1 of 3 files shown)


src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -533,4 +534,16 @@ private String getAndroidPackageName() {
         }
         return Collections.emptyList();
     }
+
+    /**
+     * Scans the classpath for Attach Services
+     * and extracts all aar libraries found
+     */
+    private void copyAarLibraries() throws IOException, InterruptedException {
+        Path libPath = getAndroidProjectPath().resolve("libs");
```
johanvos, line 543: what if this doesn't exist?

jperedadnr: That is taken care of by `FileOps.extractFilesFromJar`

PR #680, opened 2020-06-29
Use specific static libs directory in GraalVM

### LINKED ISSUE #679: Static GraalVM libraries moved to a different location
The static libraries provided by GraalVM have recently been moved to a different location. They used to be located in the `lib` folder, but are now moved to a platform dependent folder `lib/static/$platform-$arch`. New locations for each platform are:

- darwin: `lib/static/darwin-amd64`
- linux amd64: `lib/static/linux-amd64/glibc`
- linux aarch64: `lib/static/linux-aarch64/glibc`
- windows: `lib/static/windows-amd64`

tiainen:
Fixes #679

### DIFF
--- a/src/main/java/com/gluonhq/substrate/util/FileDeps.java
+++ b/src/main/java/com/gluonhq/substrate/util/FileDeps.java
@@ -87,8 +87,9 @@ public final class FileDeps {
     /**
      * Returns the path to the Java SDK static libraries for this configuration. The path is cached on the provided
      * configuration. If no custom directory has been set in the project configuration and <code>useGraalPath</code>
-     * is set to <code>true</code>, it will use the <code>lib</code> directory inside the configured Graal path. If
-     * <code>useGraalPath</code> is set to <code>false</code>, a custom Java SDK will be retrieved.
+     * is set to <code>true</code>, it will use the <code>lib/static/$target-$arch/</code> directory inside the
+     * configured Graal path. If <code>useGraalPath</code> is set to <code>false</code>, a custom Java SDK will
+     * be downloaded.
      *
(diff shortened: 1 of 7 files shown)


src/main/java/com/gluonhq/substrate/util/FileDeps.java
```diff
@@ -87,16 +87,22 @@ public FileDeps(InternalProjectConfiguration config) {
...
-     * <code>useGraalPath</code> is set to <code>false</code>, a custom Java SDK will be retrieved.
+     * is set to <code>true</code>, it will use the <code>lib/static/$target-$arch/</code> directory inside the
+     * configured Graal path. If <code>useGraalPath</code> is set to <code>false</code>, a custom Java SDK will
+     * be downloaded.
      *
      * @param useGraalPath specifies if the default Java SDK path should resolve to the Graal installation dir
      * @return the location of the static libraries of the Java SDK for the arch-os for this configuration
      * @throws IOException in case anything goes wrong.
      */
     public Path getJavaSDKLibsPath(boolean useGraalPath) throws IOException {
```
johanvos, line 98: need to return 2 paths here (the old and the new), so this is List<Path>

tiainen: Wouldn't it be better to just check for the existence of the "lib/static" directory? If it does, return that, otherwise return the old path?

johanvos: For now I'm ok with that, but in general (different issue), I feel we have to add more safety to the paths we link from (more paths in a specific order, without specifying logic)

tiainen:
Fixes #681 by specifying the library name instead of the absolute path of the static library.

PR #701, opened 2020-07-20
Use new flag for bionic

### LINKED ISSUE #700: UseBionicC removed
Follow up of https://github.com/gluonhq/client-maven-plugin/issues/211

jperedadnr:
Fixes #700

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
@@ -212,7 +212,7 @@ public class AndroidTargetConfiguration extends PosixTargetConfiguration {
         if (projectConfiguration.getGraalVersion().compareTo(new Version("20.2.0")) >= 0) {
             flags.add("--libc=bionic");
         } else if (projectConfiguration.getGraalVersion().compareTo(new Version("20.1.0")) > 0) {
-            flags.add("--H:+UseBionicC");
+            flags.add("-H:+UseBionicC");
         }
         return flags;
     }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -210,7 +210,7 @@ public boolean runUntilEnd() throws IOException, InterruptedException {
             flags.add("-H:CustomLD=" + ldlld.toAbsolutePath().toString());
         }
         if (projectConfiguration.getGraalVersion().compareTo(new Version("20.1.0")) > 0) {
-            flags.add("-H:+UseBionicC");
+            flags.add("--libc=bionic");
```
johanvos, line 213: since what version is this? If it is 20.3-dev and on, we can test for that.

jperedadnr: I believe it started with one of the last versions of 20.2.0-dev (I tested https://github.com/graalvm/graalvm-ce-dev-builds/releases/tag/20.2.0-dev-20200718_0205). 
It's safe to check for version 20.3.0 though.

gradinac: @jperedadnr You are correct, however, this flag will be included in 20.2 too, so if you are running on 20.2, it should be `--libc=bionic`

mipastgt: Yes, I reported that in https://github.com/gluonhq/client-maven-plugin/issues/211 for 20.2.0-dev-20200718_0205. Therefore it is currently not possible to build for Android.

jperedadnr: I've pushed a change to check version 20.2 and above. This will allow using the latest 20.2.0-dev version and the official release 20.2 (when released), or any 20.3-dev versions and beyond.

PR #706, opened 2020-07-23
Use ForceNoROSectionRelocations new flag

jperedadnr:


src/main/java/com/gluonhq/substrate/target/AndroidTargetConfiguration.java
```diff
@@ -202,17 +202,21 @@ public boolean runUntilEnd() throws IOException, InterruptedException {
...
-                "-H:+UseOnlyWritableBootImageHeap",
                 "-H:+UseCAPCache",
                 "-H:CAPCacheDir=" + getCapCacheDir().toAbsolutePath().toString(),
                 "-H:CompilerBackend=" + projectConfiguration.getBackend()));
         if (projectConfiguration.isUseLLVM()) {
             flags.add("-H:CustomLD=" + ldlld.toAbsolutePath().toString());
         }
-        if (projectConfiguration.getGraalVersion().compareTo(new Version("20.2.0")) >= 0) {
+        Version graalVersion = projectConfiguration.getGraalVersion();
+        if (graalVersion.compareTo(new Version("20.2.0")) >= 0) {
```
johanvos, line 212: We don't support anything lower than 20.2 I assume

jperedadnr: I guess so, but then we can remove support of old flags (like `UseBionicC`). 
Probably we should add an initial check at startup, that fails if user is using lower than 20.2?

jperedadnr: @johanvos I'd propose we integrate this PR now and do a follow-up issue to check GraalVM version is 20.2+, removing existing support for older versions.

PR #739, opened 2020-08-10
Use pathing jar to shorten command line

### LINKED ISSUE #597: "the command line is too long" exception on windows
On windows, with a lot of dependencies, the ProcessBuilder in `AbstractTargetConfiguration#compile` throws a `the command line is too long` exception on compilation.

An easy workaround is to add a `client/$app/gvm/jar` folder to the build files and copy the classpath jars here. The classpath passed to the native-image command can then filter out all jars & add `client/$app/gvm/jar/*` instead.

jperedadnr:
Fixes #597

### DIFF
--- a/src/main/java/com/gluonhq/substrate/util/FileOps.java
+++ b/src/main/java/com/gluonhq/substrate/util/FileOps.java
@@ -80,6 +80,9 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Predicate;
+import java.util.jar.Attributes;
+import java.util.jar.JarOutputStream;
+import java.util.jar.Manifest;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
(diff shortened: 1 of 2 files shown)


src/main/java/com/gluonhq/substrate/util/FileOps.java
```diff
@@ -671,6 +674,28 @@ public static void downloadAndUnzip(String sourceUrl, Path destPath, String file
...
+     * @throws IOException
+     */
+    public static String createPathingJar(String classpath) throws IOException {
+        Objects.requireNonNull(classpath);
+        Manifest manifest = new Manifest();
+        Attributes attributes = manifest.getMainAttributes();
+        attributes.put(Attributes.Name.MANIFEST_VERSION, "1.0");
+        attributes.put(Attributes.Name.CLASS_PATH,
+                classpath.replaceAll(File.pathSeparator, " "));
+        File jarFile = File.createTempFile("classpathJar", ".jar");
```
tiainen, line 691: One small remark. Wouldn't it be better to create the temporary pathing jar file in the directory provided by `ProcessPaths.getTmpPath()`?

jperedadnr: I don't see any difference?
`gvm/tmp` is mainly for the native-image's `TempDirectory` flag. If we want to have it under the gvm build folder (for inspection, for instance), probably better under the `gvm/<appname>` folder where we add the launcher files?

tiainen: I was thinking about it because the build/target directory is cleaned up more regularly then the system temp directory? In any case, the file size is negligibly small, so it's not really a big issue.

PR #740, opened 2020-08-11
reduce input touch radius to deliver missing touch events, fixes #640

### LINKED ISSUE #640: Majority of touch events is not delivered
I am trying to let the user move the content of a Pane via touch events. (Direct touch events or indirect scroll events) This works in principle but the output is more than jerky and not like what you know from other applications on Android for example. The reason is that for some reason the majority of touch events is just not delivered to the JavaFX application. Because of that any smooth movement is completely impossible. Interestingly this jerkyness stops as soon as you lift up your finger during a movement so that you only get inertia events (in case you use the scroll events). They are delivered in the usual, much more fine-grained way. Have a look at the attached log file and search for the long lines which contain "mpmediasoft". These lines mark the events that actually got delivered. The rest in between is just skipped. The number of skipped events seems to depend on the speed of your finger movement.

Samsung Galaxy Tab S4
client-maven-plugin 0.1.26
GraalVM 20.2.0-dev

[touch-log.txt](https://github.com/gluonhq/substrate/files/4705699/touch-log.txt)

AlmasB:


mipastgt:
It works. The movements are super smooth now.

mipastgt:
Is there any chance that this can be integrated? This is a very important fix for the overall user experience.

PR #818, opened 2020-10-21
Reset CoreSimulator before running tests on Mac

jperedadnr:


.github/workflows/build.yml
```diff
@@ -66,6 +66,10 @@ jobs:
           wget https://download2.gluonhq.com/substrate/javafxstaticsdk/openjfx-15-latest-ios-arm64-static.zip
           unzip openjfx-15-latest-ios-arm64-static.zip && cd $GITHUB_WORKSPACE
     
+      - name: Prepare Xcode (MacOS)
```
abhinayagarwal, line 69: A better name would be "Shutdown and Reset all Simulators (MacOS)"

jperedadnr: done

abhinayagarwal (approved):

PR #918, opened 2021-04-24
Include a version in the sysroot download link

### LINKED ISSUE #917: sysroot should get a version
When Gluon Substrate downloads the sysroot for e.g. linux-aarch64 from download2.gluonhq.com, there is no version associated with this.

### Expected Behavior
We want different versions, so that new versions of substrate might download new versions of the sysroot.

jperedadnr:
### Issue

Fixes #917 

### Progress

### DIFF
--- a/src/main/java/com/gluonhq/substrate/util/FileDeps.java
+++ b/src/main/java/com/gluonhq/substrate/util/FileDeps.java
@@ -82,7 +82,7 @@ public final class FileDeps {
             "platforms;android-29", "build-tools;29.0.3", "platform-tools",
             "extras;android;m2repository", "extras;google;m2repository", "ndk-bundle" };
 
-    private static final String ARCH_SYSROOT_URL = "https://download2.gluonhq.com/substrate/${arch}sysroot.zip";
+    private static final String ARCH_SYSROOT_URL = "https://download2.gluonhq.com/substrate/sysroot/${arch}sysroot-${version}.zip";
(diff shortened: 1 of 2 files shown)


src/main/java/com/gluonhq/substrate/util/FileDeps.java
```diff
@@ -477,7 +477,7 @@ private void fetchFromSdkManager() throws IOException, InterruptedException {
 
     private void downloadSysrootZip(String arch) throws IOException {
         Logger.logInfo("Downloading sysroot zip...");
-        String sysrootZip = Strings.substitute(ARCH_SYSROOT_URL, Map.of("arch", arch));
+        String sysrootZip = Strings.substitute(ARCH_SYSROOT_URL, Map.of("arch", arch, "version", Constants.DEFAULT_SYSROOT_VERSION));
```
johanvos, line 480: Isn't this going to download the new version if there is an old version installed?

jperedadnr: No, it won't, there is a check that looks for `~/.gluon/substrate/sysroot/aarch64` installed, and if it is there, there is no check of possible versions, since we don't include the version number.

johanvos (approved):
I suggest filing that follow-up issue for checking the version, and then this can be merged (just keeping track we don't forget about that follow-up issue)

PR #920, opened 2021-04-26
Use libm.a on Linux instead of libm.so

### LINKED ISSUE #919: use libm.a instead of libm.so on Linux Aarch64
Linking the dynamic library causes 
```
./HelloFX: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./HelloFX) 
```
when running on embedded devices with older glibc versions, like Raspbian 64 bits.

Using the static version solves the issue.

jperedadnr:
### Issue

Fixes #919 

### Progress


tiainen (approved):

johanvos (approved):
I believe libm.a requires libc-dev to be installed, but I guess that is installed whenever someone is using gcc or similar tools.
We just need to keep this into account in case there are reports about libm.a not found (libm.so typically comes with libc so no other package needed)

PR #926, opened 2021-04-27
Remove duplicated strings from link command

### LINKED ISSUE #924: Duplicated libraries in link command
The link command includes some duplicates that can be removed (from left to right).

jperedadnr:
### Issue

Fixes #924 

### Progress

### DIFF
--- a/src/main/java/com/gluonhq/substrate/util/Strings.java
+++ b/src/main/java/com/gluonhq/substrate/util/Strings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020, Gluon
+ * Copyright (c) 2019, 2021, Gluon
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


src/main/java/com/gluonhq/substrate/util/Strings.java
```diff
@@ -108,4 +112,46 @@ public static String randomString(int targetStringLength) {
                 .toString();
     }
 
+    /**
+     * Removes duplicated strings from a sourceList list, keeping the items with higher index
+     * in the list. Any item in the keep list will not be removed.
```
johanvos, line 117: that is not guaranteed to be the correct approach. If an entry appear twice, there are a number of possible scenario's. Both may be needed, first may be needed, last may be needed.

jperedadnr: There is a keep list for strings that might be duplicated and needed. 
So far the case I found is the whole/no-whole combination. 
I've tested this PR in all possible platforms, and there was no issue removing the rest of the duplicates.

PR #928, opened 2021-04-29
Fail fast if JavaFX static path is set

### LINKED ISSUE #922: fail-fast in case there are no javafx jars on the classpath
In case for some reasons the JavaFX jars are not found on the classpath, but they are expected, we should not pass control to native-image but bail immediately.

jperedadnr:
Fail fast if JavaFX static path is set but it is not valid, and also, when set, allow mapping and check files exist. 
A few more tests are included.

### Issue

Fixes #922

### Progress

### DIFF
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -87,7 +87,20 @@ public class InternalProjectConfiguration {
             getReleaseConfiguration().setSkipSigning(true);
         }
         setJavaStaticLibs(System.getProperty("javalibspath")); // this can be safely set even if null. Default will be used in that case
-        setJavaFXStaticSDK(System.getProperty("javafxsdk"));  // this can be safely set even if null. Default will be used in that case
+        String javafxStaticSdkPath = System.getenv("JAVAFX_STATIC_SDK_PATH");
+        if (javafxStaticSdkPath != null) {
+            Path lib = Path.of(javafxStaticSdkPath, "lib");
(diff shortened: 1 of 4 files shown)


eugener (approved):

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -87,7 +87,20 @@ public InternalProjectConfiguration(ProjectConfiguration config) {
...
+                throw new IllegalArgumentException("Error: " + lib + " is not a valid SDK path.\n" +
+                        "Make sure you provide a valid JAVAFX_STATIC_SDK_PATH or unset it");
+            }
+            if (!Files.exists(lib.resolve("javafx.base.jar")) ||
+                    !Files.exists(lib.resolve("javafx.graphics.jar"))) {
+                throw new IllegalArgumentException("Error: " + lib + " doesn't contain required JavaFX jars.\n" +
+                        "Make sure you provide a valid JAVAFX_STATIC_SDK_PATH or unset it");
+            }
+        }
+        setJavaFXStaticSDK(javafxStaticSdkPath);  // this can be safely set even if null. Default will be used in that case
```
johanvos, line 103: Does this avoid a checksum check?

jperedadnr: The checksum check is only done when downloading from `FileDeps::resolvePath`:

```
        if (Files.exists(Objects.requireNonNull(path))) {
              return path;                                                // -> existing path
        }
        if (!setupDependencies()) {                           // -> download and check
        ...
        }
        
```

johanvos (approved):

PR #933, opened 2021-05-13
Add sysroot to compile additional sources for Linux AArch64

### LINKED ISSUE #932: Compiling additional sources for Linux_AArch64 requires Sysroot
So far we add `sysroot` for the linking phase, but when compiling additional sources requires in too.

jperedadnr:
### Issue

Fixes #932 

### Progress

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
@@ -251,6 +251,8 @@ public class LinuxTargetConfiguration extends PosixTargetConfiguration {
 
         if (isAarch64) {
             flags.add("-DAARCH64");
+        }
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
```diff
@@ -251,6 +251,8 @@ public boolean link() throws IOException, InterruptedException {
 
         if (isAarch64) {
```
johanvos, line 252: it is actually a test on crosscompile (similar to the linkFlags)
If you're building on an AArch64 system itself, you don't have to specify a sysroot.

jperedadnr: ok, fixed

johanvos (approved):

PR #944, opened 2021-05-28
Pass down runtimeArgs

### LINKED ISSUE #943: Apply user runtimeArgs to all targets
After #942, the `runtimeArgs` list should be passed to all targets, allowing plugins to use this list. Now the runtime arguments that are not already added by Substrate or backed via System.setProperty, have to be added to the command line when running the executable.

jperedadnr:
### Issue

Fixes #943

### Progress

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2020, Gluon
+ * Copyright (c) 2019, 2021, Gluon
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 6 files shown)


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -303,11 +307,8 @@ protected boolean compileAdditionalSources()
 
         processRunner.addArg("-I" + workDir.toString());
 
-        for (String fileName : getAdditionalSourceFiles()) {
-            FileOps.copyResource(getAdditionalSourceFileLocation()  + fileName, workDir.resolve(fileName));
-            processRunner.addArg(fileName);
-        }
-        
+        processRunner.addArgs(copyAdditionalSourceFiles(workDir));
```
johanvos, line 310: This has nothing to do with passing runtimeArgs, does it?
I'm not sure the addArg(String)  is exactly the same as addArgs(List<String>) but if you tested it, I'm ok.

johanvos: Got it, I see now what it has to do with runtimeArgs

johanvos (approved):

PR #975, opened 2021-08-03
Enable iOS Simulator

### LINKED ISSUE #252: Enable iOS-sim profile
Using the iOS simulator is very convenient for getting screenshots in different resolutions required for submitting an app to the Apple store

jperedadnr:
This PR depends on a new build of GraalVM that includes the `IOS_AMD64` platform, and a new Java static build for x86_64-ios.

### Issue

Fixes #252 

### Progress


src/main/java/com/gluonhq/substrate/target/IosTargetConfiguration.java
```diff
@@ -110,23 +111,31 @@ public IosTargetConfiguration(ProcessPaths paths, InternalProjectConfiguration c
...
-        return Arrays.asList("-H:CompilerBackend=" + Constants.BACKEND_LLVM,
-                "-H:-SpawnIsolates",
+        List<String> flags = new ArrayList<>(Arrays.asList(
                 "-H:PageSize=16384",
                 "-Dsvm.targetName=iOS",
                 "-Dsvm.targetArch=" + getTargetArch(),
                 "-H:+UseCAPCache",
-                "-H:CAPCacheDir=" + getCapCacheDir().toAbsolutePath().toString());
+                "-H:CAPCacheDir=" + getCapCacheDir().toAbsolutePath().toString()));
+        if (!isSimulator()) {
```
Noisyfox, line 134: Interesting, does this imply that the reason of using LLVM backend is because of the aarch64 arch, not because of the iOS OS?

jperedadnr: It is not the arch: Android for instance has aarch64 too, but doesn't need LLVM (we did use it some time ago though). We use the LLVM backend for iOS/arm64 to create LLVM IR bitcode.
For the simulator (x86_64) this is not needed.

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -235,10 +235,14 @@ public boolean useCustomJavafxStaticLibs() {
     }
 
     public Path getDefaultJavafxStaticPath() {
+        Triplet targetTriplet = getTargetTriplet();
+        if (new Triplet(Constants.Profile.IOS_SIM).equals(targetTriplet)) {
+            targetTriplet = new Triplet(Constants.Profile.IOS);
```
johanvos, line 240: This means the arch is removed from the triplet, but it is used later at https://github.com/gluonhq/substrate/pull/975/files#diff-1e93749c0a14ab381e919971325d7c981813fdc8799bf48a1db9ee7cb641ed6bR374

Is it removed because the download doesn't deal with arch in its name?

jperedadnr: We don't remove the arch but simply create a new triplet just so we can get the arch from iOS-AArch64. 
This is required because we don't publish static JavaFX for iOS-x86_64, as we can use the AArch64 version.

johanvos: it sounds confusing we download an AArch64 artifact because that also works on AMD64.
The AMD64 artifact probably uses fat libraries? In that case, it will work, but it's very confusing. We should probably either remove the architecture from the artifact, or have 2 artifacts. But that is tbd in CI.

johanvos (approved):

PR #1066, opened 2021-12-14
Add support for DMG bundles on macOS

### LINKED ISSUE #1061: Add support for macOS dmg bundle
Following up on #1054, add support for dmg bundles

jperedadnr:
### Issue

Fixes #1061

### Progress


src/main/java/com/gluonhq/substrate/target/MacOSTargetConfiguration.java
```diff
@@ -143,13 +143,30 @@ String getAdditionalSourceFileLocation() {
...
     public boolean packageApp() throws IOException, InterruptedException {
-        createAppBundle();
+        boolean sign = !projectConfiguration.getReleaseConfiguration().isSkipSigning();
+        createAppBundle(sign);
+
+        String packageTypes = projectConfiguration.getReleaseConfiguration().getPackageTypes();
+        if (packageTypes == null || packageTypes.isEmpty()) {
+            // if it is not set, do nothing
+            return true;
+        } else if (!(packageTypes.contains("pkg") || packageTypes.contains("dmg"))) {
```
erwin1, line 153: shouldn't this be seen as a configuration error, hence throw an error?

src/main/java/com/gluonhq/substrate/target/MacOSTargetConfiguration.java
```diff
@@ -143,13 +143,30 @@ String getAdditionalSourceFileLocation() {
...
+            return true;
+        }
 
+        boolean result = true;
         Packager packager = new Packager(paths, projectConfiguration);
-        return packager.createPackage(!projectConfiguration.getReleaseConfiguration().isSkipSigning());
+        if (packageTypes.contains("pkg")) {
+            result = packager.createPackage(sign);
+        }
+        if (result && packageTypes.contains("dmg")) {
```
erwin1, line 163: this is the reason I would have preferred to allow only one type: pkg (either standalone or appstore) or dmg

src/main/java/com/gluonhq/substrate/target/MacOSTargetConfiguration.java
jperedadnr: Since `packageTypes` can contain other types (for instance for Win or Linux), without the need of a profile, I was thinking that this shouldn't fail with an error.
But (and related to your other comment), it might be better to refactor this and force a single package type per run, and enforce the use of profiles. Anyway even `pkg` and `dmg` might need different properties that won't fit as they are now (version code, ...).

src/main/java/com/gluonhq/substrate/target/MacOSTargetConfiguration.java
jperedadnr: I'll refactor it

johanvos (approved):

erwin1 (approved):

PR #1068, opened 2021-12-15
Create Installer for Windows

abhinayagarwal:
### Issue

Fixes # 1067

The PR wraps the native executable in a MSI installer which installs and creates the start menu and desktop shortcuts.


src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
```diff
@@ -258,6 +271,22 @@ public void setAppLabel(String appLabel) {
         this.appLabel = appLabel;
     }
 
+    public String getAppDescription() {
+        return appDescription == null ? "" : appDescription;
```
jperedadnr, line 275: Why do we need an empty string, probably null is good enough as default?

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,314 @@
...
+import java.io.Writer;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.text.MessageFormat;
+import java.util.*;
```
jperedadnr, line 49: Avoid wildcards

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,314 @@
...
+        /**
+         * Create directories and copy resources
+         */
+        Path tmpMSI = paths.getTmpPath().resolve("tmpMSI");
+        if (Files.exists(tmpMSI)) {
+            FileOps.deleteDirectory(tmpMSI);
+        }
+        Files.createDirectories(tmpMSI);
+
+        Path userAssets = rootPath.resolve(Constants.MACOS_ASSETS_FOLDER);
```
jperedadnr, line 85: Use `WIN_ASSETS_FOLDER` or refactor constants to `ASSETS_FOLDER`?

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,314 @@
...
+    private final InternalProjectConfiguration projectConfiguration;
+    private final String sourceOS;
+    private final Path rootPath;
+
+    public MSIBundler(ProcessPaths paths, InternalProjectConfiguration projectConfiguration) {
+        this.paths = paths;
+        this.projectConfiguration = projectConfiguration;
+        this.sourceOS = projectConfiguration.getTargetTriplet().getOs();
+        this.rootPath = paths.getSourcePath().resolve(sourceOS);
+    }
```
jperedadnr, line 65: We should try to find WiX installer as soon as possible and fail fast if it is not found.
Also an indication on how to install it would be good.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,314 @@
...
+        LIGHT;
+
+        String getPath() {
+            for (var dir : findWixInstallDirs()) {
+                Path path = dir.resolve(name().toLowerCase() + ".exe");
+                if (Files.exists(path)) {
+                    return path.toString();
+                }
+            }
+            throw new RuntimeException("Path not found");
```
jperedadnr, line 265: Add a more descriptive message. Probably use `Logger.logSevere`. You could also add instructions on how to install it.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+
+    enum WixTool {
+        CANDLE,
+        LIGHT;
+
+        String getPath() {
+            for (var dir : findWixInstallDirs()) {
+                Path path = dir.resolve(name().toLowerCase() + ".exe");
```
jperedadnr, line 260: `toLowerCase(Locale.ROOT)`

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,314 @@
...
+        }
+        Files.createDirectories(tmpMSI);
+
+        Path userAssets = rootPath.resolve(Constants.MACOS_ASSETS_FOLDER);
+        // copy assets to gensrc/windows
+        Path windowsGenSrcPath = paths.getGenPath().resolve(sourceOS);
+        Path windowsAssetPath = windowsGenSrcPath.resolve(Constants.MACOS_ASSETS_FOLDER);
+        Files.createDirectories(windowsAssetPath);
+        // Copy system resources and over-write it with user provided resources
+        FileOps.copyDirectoryFromResources("/native/windows/assets", windowsAssetPath);
```
jperedadnr, line 91: We log a message to the user in case default resources are used, with instructions on how to overwrite them.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: Changed it to `WIN_ASSETS_FOLDER`. Hopefully, we will replace all the redundant constants with `ASSETS_FOLDER`.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: All I found in [Packager.java](https://github.com/gluonhq/substrate/blob/master/src/main/java/com/gluonhq/substrate/util/macos/Packager.java#L183) was instruction to copy the generated file. Can you point me to a location where we instruct user on how to overwrite defaults?

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
abhinayagarwal: If we return null from the configuration, we need to add a null check in MSIBundler instead. It should be the responsibility of the Data class to return the correct data.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: I have added a log with instructions.

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
jperedadnr: That would apply to the rest of the fields of this class then. But this is the only case. And I don't see a problem doing null checks?

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
jperedadnr: That same link you have posted logs: 
```
"Default background image generated in " + macosAssets.toString() + ".\n" +
                    "Consider copying it to " + rootPath.toString() + " before performing any modification"
```

src/main/java/com/gluonhq/substrate/Constants.java
```diff
@@ -195,4 +195,8 @@
      */
     public static final String NATIVE_IMAGE_ARG_CLASSPATH = "-cp";
 
+    /**
+     * Pacakge
```
jperedadnr, line 199: Typo

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: The message seems misdirected. I feel its asking me to copy the default file to the suggested path before I can make any modification.

jperedadnr: The default assets are copied to a target location that gets removed, so any change to them at that location can get lost. We ask developers to copy to a src location, keeping path and name of asset, and then there they can do whatever modification is needed. Feel free to reword, but this is what we are asking.

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
abhinayagarwal: There isn't an issue with null checks. However, there are some methods like `getBundleVersion` which returns null and then we check for the value and assign a default value which is fetched from parameter placed in the same class `ReleaseConfiguration#DEFAULT_BUNDLE_VERSION`.

I feel that ReleaseConfiguration methods can take care of these null checks and return default values, if required. But, as you said, it will require change to the entire class. I will update the code as you suggested for now.

src/main/java/com/gluonhq/substrate/Constants.java
```diff
@@ -195,4 +195,8 @@
      */
     public static final String NATIVE_IMAGE_ARG_CLASSPATH = "-cp";
 
+    /**
+     * Package
+     */
+    public static final String DEFAULT_APP_VERSION = "1.0.0";
```
jperedadnr, line 201: The package constants are at ReleaseConfiguration by now. They should be all together at a single place. Move them here?

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
```diff
@@ -126,6 +126,13 @@
      */
     private String appLabel;
 
+    /**
+     * A short description about the application
+     *
+     * Default: Empty string.
+     */
+    private String appDescription;
```
jperedadnr, line 134: These two don't belong to Android. If they are Win specific, add a comment `// Windows'  before them. If they are valid for all platforms, put them before the Mac ones (I see they are).

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
```diff
@@ -258,6 +271,22 @@ public void setAppLabel(String appLabel) {
...
+    }
+
+    public String getVendor() {
+        return vendor;
+    }
+
+    public void setVendor(String vendor) {
+        this.vendor = vendor;
+    }
+
```
jperedadnr, line 289: Add them to `toString()`.

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
abhinayagarwal: Currently these are windows specific, but I can foresee these being used in other platforms as well.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+                WixTool.CANDLE.getPath(),
+                "-nologo",
+                wixPath.toString(),
+                "-ext", "WixUtilExtension",
+                "-arch", "x64",
+                "-out", wixObjPath.toString()
+        ));
+
+        userInput.entrySet().stream()
+                .map(wixVar -> String.format("-d%s=%s", wixVar.getKey(), wixVar.getValue()))
```
jperedadnr, line 128: While testing I've noticed that `appName` might contain spaces, and that fails.
Use quotes for values.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
jperedadnr: Still pending

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+        Path msiPath = paths.getGvmPath().getParent().resolve(appName + "-" +  version  + ".msi");
+        FileOps.copyResource("/native/windows/wix/main.wxs", wixPath);
+        FileOps.copyDirectory(windowsAssetPath, config);
+
+        /**
+         * Wix Compile
+         */
+        Map<String, String> userInput = createAppDetailMap();
+        List<String> processArgs = new ArrayList<>(List.of(
+                WixTool.CANDLE.getPath(),
```
jperedadnr, line 119: The paths in this command need to be wrapped with quotes.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+        if (candle.runProcess("Wix Compiler") != 0) {
+            throw new IOException("Error running candle to generate wixobj");
+        }
+
+        /**
+         * Wix Link
+         */
+        processArgs.clear();
+        processArgs.addAll(List.of(
+                WixTool.LIGHT.getPath(),
```
jperedadnr, line 141: Same here

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+
+        Map<String, String> userInput = new HashMap<>();
+        ReleaseConfiguration releaseConfiguration = projectConfiguration.getReleaseConfiguration();
+
+        String appName = projectConfiguration.getAppName();
+        String executableName = appName + ".exe";
+        String vendor = Optional.ofNullable(releaseConfiguration.getVendor()).orElse("Unknown");
+        String version = Optional.ofNullable(releaseConfiguration.getVersionName()).orElse(DEFAULT_APP_VERSION);
+        userInput.put("GSProductCode", createUUID("ProductCode", appName, vendor, version).toString());
+        userInput.put("GSAppName", appName);
```
jperedadnr, line 169: If `appName` contains spaces, `<File Id="$(var.GSAppName)"` fails with:
```
error CNDL0014 : The File/@Id attribute's value, '"My Gluon Application"', is not a legal identifier.  Identifiers may contain ASCII characters A-Z, a-z, digits, underscores (_), or periods (.).  Every identifier must begin with either a letter or an underscore.
```
Maybe you need some `GSAppNameId` where you use `appName.replaceAll("\\s","_")`

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+        userInput.put("GSAppExecutable", executableName);
+        userInput.put("GSAppVersion", version);
+        userInput.put("GSAppVendor", vendor);
+        userInput.put("GSAppIconName", appName + ".ico");
```
jperedadnr, line 173: Same here, add `GSAppIconNameId`.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+                "-ext", "WixUtilExtension",
+                "-arch", "x64",
+                "-out", wixObjPath.toString()
+        ));
+
+        userInput.entrySet().stream()
+                .map(wixVar -> String.format("-d%s=%s", wixVar.getKey(), wixVar.getValue()))
+                .forEachOrdered(processArgs::add);
+        Logger.logInfo(String.join(" ", processArgs));
+        ProcessRunner candle = new ProcessRunner(processArgs.toArray(new String[0]));
```
jperedadnr, line 131: With the changes I've mentioned before, I get a valid command line, but the process still fails:
```
Output 
====== 
candle.exe : error CNDL0103 : The system cannot find the file 'Gluon' with type 'Source'.   
Result
 ====== 
result: 103
```
However if I copy that command line from the log and paste it directly on a terminal it works fine.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: Isn't it the responsibility of the user to provide the correct app name?

jperedadnr: hmm, we use for the app name the `<name/>` from the Maven [project](https://github.com/gluonhq/gluonfx-maven-plugin/blob/master/src/main/java/com/gluonhq/NativeBaseMojo.java#L202), which is "the full name of the project", so I take it can have spaces.

Dealing with this is cumbersome but, unless we force a check on the appName to not contain spaces, we must take it into account in Substrate.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+                wixPath.toString(),
+                "-ext", "WixUtilExtension",
+                "-arch", "x64",
+                "-out", wixObjPath.toString()
+        ));
+
+        userInput.entrySet().stream()
+                .map(wixVar -> String.format("-d%s=%s", wixVar.getKey(), wixVar.getValue()))
+                .forEachOrdered(processArgs::add);
+        Logger.logInfo(String.join(" ", processArgs));
```
jperedadnr, line 130: Move to `logDebug` or remove (the ProcessRunner should be logged either way)

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+                "-ext", "WixUIExtension",
+                "-out", msiPath.toString(),
+                wixObjPath.toString()
+        ));
+
+        ProcessRunner light = new ProcessRunner(processArgs.toArray(new String[0]));
+        if (light.runProcess("Wix Linker") != 0) {
+            throw new IOException("Error running light to generate msi");
+        }
+
```
jperedadnr, line 155: Add a final message like in https://github.com/gluonhq/substrate/blob/master/src/main/java/com/gluonhq/substrate/util/macos/Packager.java#L243

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,325 @@
...
+        userInput.put("GSAppIcon", paths.getTmpPath().resolve("tmpMSI").resolve("config").resolve("icon.ico").toString());
+        userInput.put("GSMainExecutableGUID", createUUID("GSMainExecutableGUID", appName, vendor, version).toString());
+        userInput.put("GSStartMenuShortcutGUID", createUUID("StartMenuShortcutGUID", appName, vendor, version).toString());
+        userInput.put("GSDesktopShortcutGUID", createUUID("DesktopShortcutGUID", appName, vendor, version).toString());
+        Path license = paths.getTmpPath().resolve("tmpMSI").resolve("config").resolve("license.rtf");
+        if (Files.exists(license)) {
+            ensureByMutationFileIsRTF(license);
+            userInput.put("GSLicenseRtf", license.toString());
+        }
+        userInput.put("GSApplicationPath", paths.getClientPath().resolve("x86_64-windows").resolve(executableName).toString());
```
jperedadnr, line 183: Replace `paths.getClientPath().resolve("x86_64-windows")` with `paths.getAppPath()`

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
jperedadnr: See Linux case, for instance. They apply some conditions: https://github.com/openjdk/jdk/blob/master/src/jdk.jpackage/linux/classes/jdk/jpackage/internal/LinuxDebBundler.java#L79 and create an installer without spaces. We could do the same (copying the native image binary to tmpMSI and changing the name accordingly).

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: Do you have a reproducible sample for this?

jperedadnr: I’m using a project generated from start.gluon.io, but any project with `<name>My Application</name>` in its pom will do.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: I have added functionality to trim app name of spaces before it is passed to Wix toolset

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: I tried out this command and it doesn't fail for paths which contains space in it:

```
C:\Program Files (x86)\WiX Toolset v3.11\bin\candle.exe ...
```

The above path just worked fine.

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: same as above

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: Fixed this with recent commit

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
abhinayagarwal: Fixed with recent commit

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,327 @@
...
+        // copy assets to gensrc/windows
+        Path windowsGenSrcPath = paths.getGenPath().resolve(sourceOS);
+        Path windowsAssetPath = windowsGenSrcPath.resolve(Constants.WIN_ASSETS_FOLDER);
+        Files.createDirectories(windowsAssetPath);
+        // Copy system resources and over-write it with user provided resources
+        FileOps.copyDirectoryFromResources("/native/windows/assets", windowsAssetPath);
+        Logger.logInfo("Default icon.ico image generated in " + windowsAssetPath + ".\n" +
+                "Consider copying it to " + rootPath + " before performing any modification");
+        if (Files.exists(userAssets)) {
+            FileOps.copyDirectory(userAssets, windowsAssetPath);
```
jperedadnr, line 103: This works, but this is not how we do it for the resources and assets of other platforms.
We don't copy user assets into gensrc, which is only for default resources. 

In the next step you could simply do:
```
if (Files.exists(userAssets)) {
    FileOps.copyDirectory(userAssets, config);
} else {
    FileOps.copyDirectory(windowsAssetPath, config);
}
```

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
jperedadnr: Tested it, looks good.

jperedadnr (commented):
Looks good, two minor comments pending

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,328 @@
...
+        Path userAssets = rootPath.resolve(Constants.WIN_ASSETS_FOLDER);
+        // copy assets to gensrc/windows
+        Path windowsGenSrcPath = paths.getGenPath().resolve(sourceOS);
+        Path windowsAssetPath = windowsGenSrcPath.resolve(Constants.WIN_ASSETS_FOLDER);
+        Files.createDirectories(windowsAssetPath);
+        Path config = tmpMSI.resolve("config");
+        Files.createDirectories(config);
+        // Copy system resources and over-write it with user provided resources
+        FileOps.copyDirectoryFromResources("/native/windows/assets", windowsAssetPath);
+        Logger.logInfo("Default icon.ico image generated in " + windowsAssetPath + ".\n" +
```
jperedadnr, line 102: move this message to the else clause? It makes sense only when users don't have their own resources

jperedadnr (approved):
Looks good. It needs a follow-up on GluonFX maven plugin for ReleaseConfiguration changes.

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
```diff
@@ -36,6 +36,8 @@
     public static final String DEFAULT_CODE_VERSION = "1";
     public static final String DEFAULT_CODE_NAME = "1.0";
 
+    public static final String DEFAULT_APP_VERSION = "1.0.0";
```
jperedadnr, line 39: Update copyright year

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
```diff
@@ -275,6 +290,22 @@ public void setAppLabel(String appLabel) {
         this.appLabel = appLabel;
     }
 
+    public String getAppDescription() {
```
jperedadnr, line 293: Move up too, right after get/setPackageType

src/main/java/com/gluonhq/substrate/model/ReleaseConfiguration.java
```diff
@@ -338,6 +369,8 @@ public String toString() {
                 ", skipSigning=" + skipSigning +
                 ", simulatorDevice='" + simulatorDevice + '\'' +
                 ", appLabel='" + appLabel + '\'' +
+                ", appDescription='" + appDescription + '\'' +
```
jperedadnr, line 372: Move up, right after packageType

jperedadnr (approved):

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,343 @@
...
+
+        Path userAssets = rootPath.resolve(Constants.WIN_ASSETS_FOLDER);
+        Path iconPath = null;
+        if (Files.exists(userAssets) && Files.isDirectory(userAssets)) {
+            iconPath = userAssets.resolve("icon.ico");
+        }
+
+        // copy assets to gensrc/macos
+        if (iconPath == null || !Files.exists(iconPath)) {
+            iconPath = FileOps.copyResource("/native/windows/assets/icon.ico", iconPath);
```
jperedadnr, line 115: the second `iconPath` is null, use: `windowsAssetPath.resolve("icon.ico")` instead

src/main/java/com/gluonhq/substrate/util/windows/MSIBundler.java
```diff
@@ -0,0 +1,343 @@
...
+         * Create directories and copy resources
+         */
+        Path tmpMSI = paths.getTmpPath().resolve("tmpMSI");
+        if (Files.exists(tmpMSI)) {
+            FileOps.deleteDirectory(tmpMSI);
+        }
+        Files.createDirectories(tmpMSI);
+
+        Path windowsGenSrcPath = paths.getGenPath().resolve(sourceOS);
+        Path windowsAssetPath = windowsGenSrcPath.resolve(Constants.WIN_ASSETS_FOLDER);
```
jperedadnr, line 102: This can be moved below, since it is only used if (iconPath == null || !Files.exists(iconPath)):

```
if (iconPath == null || !Files.exists(iconPath)) {
    Path windowsGenSrcPath = paths.getGenPath().resolve(sourceOS);
    Path windowsAssetPath = windowsGenSrcPath.resolve(Constants.WIN_ASSETS_FOLDER);
    Files.createDirectories(windowsAssetPath);
...
}

abhinayagarwal: Should I also move creation of tmpMSI and config before the icon is copied to config directory?

jperedadnr: That is not required, it has to be done anyway.

jperedadnr (approved):

PR #1074, opened 2021-12-20
Fix sunec check

### LINKED ISSUE #1072: Linux still tries to link the sunec library under JDK 17
After #1028, `sunec` shouldn't be added to the link command using JDK 17. This fails on Linux:

AbstractTargetConfiguration gets `sunec` from `defaultStaticJavaLibs`, subclasses like MacOSTargetConfiguration, WindowsTargetConfiguration and LinuxTargetConfiguration have also `staticJavaLibs` with `sunec`.

However, LinuxTargetConfiguration processes that list and returns `libsunec.a`. The check in AbstractTargetConfiguration::getTargetSpecificJavaLinkLibraries fails for Linux.

jperedadnr:
### Issue

Fixes #1072 

### Progress


johanvos (approved):
temporary fix is ok, we need to deal with libsunec (and others) much earlier in the process though.

PR #1100, opened 2022-01-28
Run `ie4uinit.exe -show` in Windows Server 2019

### LINKED ISSUE #1099: Link step hangs indefinitely in clearExporerCache
On Windows, the link step fails when calling the `ie4uinit.exe` in the `clearExplorerCache()` method:

https://github.com/gluonhq/substrate/blob/master/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java#L238

When inspecting task manager, the process is indeed active, and ending it causes the build to continue.

The reason is that the build is running on Windows Server 2019 and the `os.name` system property is not correctly checked in the above line. This makes substrate believe that it is running on a Windows whose version is less than 10.

The value of the `os.name` system property is `Windows Server 2019`. Adding that check so that substrate also runs `ie4uinit.exe -show` instead of `ie4uinit.exe -ClearIconCache` fixes the issue.

abhinayagarwal:
Fixes #1099

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
@@ -291,7 +291,7 @@ public class WindowsTargetConfiguration extends AbstractTargetConfiguration {
             Logger.logDebug("Windows Build number: " + buildNumber);
             flag = Integer.parseInt(buildNumber) > 10000 ? "-show" : "-ClearIconCache";
         } catch (Exception e) {
-            // do nothing
+            Logger.logInfo("Unable to find Windows build version. Defaulting cache flag to '-show'.");
         }
         return flag;
     }
--- a/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
@@ -42,7 +42,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.Locale;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
```diff
@@ -235,7 +235,7 @@ private void clearExplorerCache() throws IOException, InterruptedException {
         ProcessRunner clearCache = new ProcessRunner("ie4uinit");
         // For Windows 10 and later, use `ie4uinit.exe -show`
         // For Windows version < 10, use `ie4uinit.exe -ClearIconCache`
-        clearCache.addArg(osName.equals("windows 10") || osName.equals("windows 11") ? "-show" : "-ClearIconCache");
+        clearCache.addArg(osName.equals("windows 10") || osName.equals("windows 11") || osName.equals("windows server 2019") ? "-show" : "-ClearIconCache");
```
jperedadnr, line 238: It could probably better to have a default set to `-show`, and only if windows version < 10 set the other option. The problem is finding out which versions are those (Windows 7, 8, 8.1, server versions 2016? 2012...)

Another option: use the Windows build number. Checking https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions, it seems build numbers > 10000 could work.

Running `ver` should give you either major version and build version
```
$ ver
Microsoft Windows [Version 10.0.14366]
```

tiainen (approved):

src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
```diff
@@ -279,4 +267,22 @@ private void createIconResource() throws InterruptedException, IOException {
...
+    private void clearExplorerCache() throws IOException, InterruptedException {
+        ProcessRunner clearCache = new ProcessRunner("ie4uinit");
+        clearCache.addArg(findCacheFlag());
+        clearCache.runProcess("Clear Explorer cache");
+    }
+
+    // For Windows build > 10000, use `ie4uinit.exe -show`
+    // For Windows build < 10000, use `ie4uinit.exe -ClearIconCache`
+    private String findCacheFlag() throws IOException, InterruptedException {
+        ProcessRunner buildNumberProcess = new ProcessRunner("powershell.exe", "(Get-ComputerInfo).OsBuildNumber");
```
jperedadnr, line 282: Will this work in all Windows versions? Tried on Windows 7 and the command `Get-ComputerInfo` is not recognised.
Maybe you can put it under try-catch, and if it fails we can assume it is old enough to be < 10000?

abhinayagarwal: What exception is thrown on Windows 7?

tiainen: The [`Get-ComputerInfo` cmdlet](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-computerinfo?view=powershell-7.2#description) was introduced in PowerShell 5.1, so it's possible that have an older version?

abhinayagarwal:
It looks that the command is not recognised on the system. Can you dump the specs of this PC?

abhinayagarwal:
I have pushed a fix for this. Although, this makes me wonder if we should default to `-ClearIconCache` instead of `-show` for corner cases like these?

jperedadnr:
We need a follow-up PR in order to cache this version number (to ./gluon/substrate for instance) and avoid doing the check on every link.

jperedadnr (approved):

PR #1105, opened 2022-02-02
Check GraalVM vendor

### LINKED ISSUE #1102: Inet6Address::cached_scoped_id removed in JDK 11.0.14
See https://bugs.openjdk.java.net/browse/JDK-8216417

The jniconfig-java11.json file includes the method `Inet6Address::cached_scoped_id`, but that can only be applied only until 11.0.13. Based on GraalVM's java version, Substrate will need to remove it, if JDK >= 11.0.14.

### LINKED ISSUE #1104: Detect and warn if GraalVM build is not the required version
Gluon builds of GraalVM are required to use Substrate and GluonFX plugins, however, official builds (release/dev) could be used without any guarantee.
Substrate should check the GraalVM version and warn the user about this.

jperedadnr:
### Issue

Fixes #1104
Fixes #1102

### Progress

### DIFF
--- a/src/main/java/com/gluonhq/substrate/Constants.java
+++ b/src/main/java/com/gluonhq/substrate/Constants.java
@@ -117,7 +117,7 @@ public class Constants {
     /**
      * Supported GraalVM versions
      */
-    public static final String GRAALVM_MIN_VERSION = "21.0";
+    public static final String GRAALVM_MIN_VERSION = "22.0";
     public static final String GRAALVM_JAVA_MIN_VERSION = "11.0";
 
     /**
--- a/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
+++ b/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
@@ -380,6 +380,7 @@ public class SubstrateDispatcher {
 
         this.config.checkGraalVMVersion();
         this.config.checkGraalVMJavaVersion();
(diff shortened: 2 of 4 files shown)


src/main/java/com/gluonhq/substrate/Constants.java
```diff
@@ -117,7 +117,7 @@
     /**
      * Supported GraalVM versions
      */
-    public static final String GRAALVM_MIN_VERSION = "21.0";
+    public static final String GRAALVM_MIN_VERSION = "22.0";
```
johanvos, line 120: shouldn't this be 22.0.0.3 since the cached_scope_id field is removed?

jperedadnr: Hmm, we only check major.minor.patch versions, but it could be added

johanvos: not really needed though with the check on the vendor which is part of this PR

johanvos (approved):

tiainen (approved):

src/main/java/com/gluonhq/substrate/Constants.java
jperedadnr: The check will log a warning only, so those using 22.0.0.2 or lower will need to add the method to the jni list or use runAgent

PR #1184, opened 2022-10-31
Change default value from 1 to 2

### LINKED ISSUE #1183: Change -Dmonocle.input.touchRadius value to 2
Following https://stackoverflow.com/questions/74097583/combobox-makes-selection-with-touch-events-difficult, a value of 1 causes touch events to not register correctly on some devices (tested on 2 Samsung Galaxys). A value of 2 works better. I suggest changing the default for Andoird to 2 at https://github.com/gluonhq/substrate/blob/master/src/main/resources/native/android/c/launcher.c#L69.

nlisker:
Changes the value of `-Dmonocle.input.touchRadius` from 1 to 2.

Fixes #1183

### Progress

### DIFF
--- a/src/main/resources/native/android/c/launcher.c
+++ b/src/main/resources/native/android/c/launcher.c
@@ -66,7 +66,7 @@ const char *origArgs[] = {
     "-Dcom.sun.javafx.gestures.rotate=true",
     "-Dcom.sun.javafx.gestures.scroll=true",
     "-Djavafx.verbose=false",
-    "-Dmonocle.input.touchRadius=1",
+    "-Dmonocle.input.touchRadius=2",
     "-Dmonocle.input.traceEvents.verbose=false",
     "-Dprism.verbose=true",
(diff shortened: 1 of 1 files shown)


javateer:
Do not do this.

Changing the default value from 1 to 2 imposes that change onto all developers' applications. This change comes at the cost of scrolling lists no longer moving in a smooth, fluid manner. Instead, the widgets being scrolled shift in a choppy / shakey undesirable fashion. This being "the norm" of an app built using GluonFX will hurt its brand.

An appropriate fix with demonstration code provided [here](https://github.com/gluonhq/substrate/pull/1183) proves the default value of monocle.input.touchRadius = 1 can remain, and still, the GluonFX touch/drag events work as expected across all devices, including the problematic Samsung.

Anyone that still has cause to want to increase monocle.input.touchRadius to value of 2 or greater can still do so onto their own application without burdening the entire community to override that change back to 1 in their own apps' configuration.

PR #1209, opened 2023-03-04
Let Android OS handle volume events

### LINKED ISSUE #1208: [Android] KeyEvents from volume buttons shouldn't be consumed
Android MainActivity currently consumes all the events after passing them to the JavaFX layer. This applies to the volume (Up/Down) events from the volume buttons, which prevents from getting the system volume menu to show up and adjust the volume level accordingly. 
This implies that the event has to be managed from the JavaFX layer instead (sending a callback to Android to process the event).

See https://github.com/gluonhq/attach/pull/348/commits/f2025d84eefb2be939790a10f996d0a7a839f5d8 for reference.

jperedadnr:
### Issue

Fixes #1208

### Progress

### DIFF
--- a/src/main/resources/native/android/android_project/app/src/main/java/com/gluonhq/helloandroid/MainActivity.java
+++ b/src/main/resources/native/android/android_project/app/src/main/java/com/gluonhq/helloandroid/MainActivity.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Gluon
+ * Copyright (c) 2019, 2023, Gluon
  *
  * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


salmonb:
I tested this PR along with my [PR](https://github.com/gluonhq/attach/pull/348) regarding the AndroidAudioService, and my tests confirm that this PR is fixing the issue with the volume buttons: now they are simply handled by Android, without the need to rewrite the code for the volume management in the service (calling `activity.setVolumeControlStream()` is enough to make the volume buttons work) 👍

jperedadnr:
Great, thanks for testing!

johanvos (approved):

PR #1234, opened 2023-11-07
added support for graalvm version strings without minor version

ennerf:
Substrate currently fails for some GraalVM builds that print versions without minor revision numbers, e.g., `graalvm-jdk-21+35.1/bin/java -version`

```bash
java version "21" 2023-09-19
Java(TM) SE Runtime Environment Oracle GraalVM 21+35.1 (build 21+35-jvmci-23.1-b15)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 21+35.1 (build 21+35-jvmci-23.1-b15, mixed mode, sharing)
```

This PR changes the Regex from `GraalVM .*?(\d\d.\d.\d)` to  `GraalVM (\d{1,2}(\.\d+){0,2})` (same as for the Java version) and adds unit tests for a variety of old and new GraalVM builds on different platforms 

### Current Regex

[GraalVM .*?(\d\d.\d.\d)](https://regex101.com/r/CPZMHR/1)

image

### Updated Regex

[GraalVM (\d{1,2}(\.\d+){0,2})](https://regex101.com/r/DHE4ti/2)

image

- [ ] Change must not contain extraneous whitespace
- [ ] License header year is updated, if required
- [x] Verify the contributor has signed [Gluon Individual Contributor License Agreement (CLA)](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)

PR #1236, opened 2023-11-11
Add GraalVM for JDK 21 support.

kkriske:
- no longer register HomeFinderFeature post JDK 21
- no longer link against fdlibm on Windows and Linux post JDK 21
- link against extnet and mswsock on Windows post JDK 21
- add `--add-exports` flag required for the GluonFeature internal API usage

### Issue

Builds against GraalVM for JDK 21 run into some issues:
* The HomeFinderFeature is no longer available by default since it's part of polyglot, and polyglot has been split off to separate maven artifacts. The feature is also automatically registered in a native-image.properties file in the META-INF of that dependency so it should not be required to register it manually.

Windows specific issues:
* the `nio` library now depends on `mswsock`, so that needs to be added to the linker
* the `fdlibm` library is no longer part of the JDK so it should be removed from the linker, not sure since what exact JDK version
* the `extnet` library is now used on Windows in the `WindowsSocketOptions` class, so it needs to be added to the linker

Linux specific issues:
* the `fdlibm` library is no longer part of the JDK so it should be removed from the linker, not sure since what exact JDK version

Other:
* `fdlibm` should probably be removed from the remaining targets as well, but I have no way to verify that.

### Progress

- [x] Verify the contributor has signed [Gluon Individual Contributor License Agreement (CLA)](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)

### DIFF
--- a/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
+++ b/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
@@ -379,29 +379,25 @@ public class SubstrateDispatcher {
             System.out.println("Configuration: " + this.config);
         }
 
-        Version javaVersion = this.config.checkGraalVMJavaVersion();
-        this.config.checkGraalVMVersion(javaVersion);
-        this.config.checkGraalVMVendor();
-
         Triplet targetTriplet = config.getTargetTriplet();
 
-        this.targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet, javaVersion),
+        this.targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet),
                 "Error: Target Configuration was not found for " + targetTriplet);
 
         Logger.logInit(paths.getLogPath().toString(), this.config.isVerbose());
     }
 
-    private TargetConfiguration getTargetConfiguration(Triplet targetTriplet, Version javaVersion) throws IOException {
+    private TargetConfiguration getTargetConfiguration(Triplet targetTriplet) throws IOException {
         if (!Constants.OS_WEB.equals(targetTriplet.getOs()) && !config.getHostTriplet().canCompileTo(targetTriplet)) {
             throw new IllegalArgumentException("We currently can't compile to " + targetTriplet + " when running on " + config.getHostTriplet());
         }
         switch (targetTriplet.getOs()) {
-            case Constants.OS_LINUX  : return new LinuxTargetConfiguration(paths, config, javaVersion);
-            case Constants.OS_DARWIN : return new MacOSTargetConfiguration(paths, config, javaVersion);
-            case Constants.OS_WINDOWS: return new WindowsTargetConfiguration(paths, config, javaVersion);
-            case Constants.OS_IOS    : return new IosTargetConfiguration(paths, config, javaVersion);
-            case Constants.OS_ANDROID: return new AndroidTargetConfiguration(paths, config, javaVersion);
-            case Constants.OS_WEB    : return new WebTargetConfiguration(paths, config, javaVersion);
+            case Constants.OS_LINUX  : return new LinuxTargetConfiguration(paths, config);
+            case Constants.OS_DARWIN : return new MacOSTargetConfiguration(paths, config);
+            case Constants.OS_WINDOWS: return new WindowsTargetConfiguration(paths, config);
+            case Constants.OS_IOS    : return new IosTargetConfiguration(paths, config);
+            case Constants.OS_ANDROID: return new AndroidTargetConfiguration(paths, config);
+            case Constants.OS_WEB    : return new WebTargetConfiguration(paths, config);
             default                  : return null;
         }
     }
--- a/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
+++ b/src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
@@ -385,23 +385,23 @@ public class SubstrateDispatcher {
 
         Triplet targetTriplet = config.getTargetTriplet();
 
-        this.targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet),
+        this.targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet, javaVersion),
                 "Error: Target Configuration was not found for " + targetTriplet);
 
         Logger.logInit(paths.getLogPath().toString(), this.config.isVerbose());
     }
 
-    private TargetConfiguration getTargetConfiguration(Triplet targetTriplet) throws IOException {
+    private TargetConfiguration getTargetConfiguration(Triplet targetTriplet, Version javaVersion) throws IOException {
         if (!Constants.OS_WEB.equals(targetTriplet.getOs()) && !config.getHostTriplet().canCompileTo(targetTriplet)) {
             throw new IllegalArgumentException("We currently can't compile to " + targetTriplet + " when running on " + config.getHostTriplet());
         }
         switch (targetTriplet.getOs()) {
-            case Constants.OS_LINUX  : return new LinuxTargetConfiguration(paths, config);
-            case Constants.OS_DARWIN : return new MacOSTargetConfiguration(paths, config);
-            case Constants.OS_WINDOWS: return new WindowsTargetConfiguration(paths, config);
-            case Constants.OS_IOS    : return new IosTargetConfiguration(paths, config);
-            case Constants.OS_ANDROID: return new AndroidTargetConfiguration(paths, config);
-            case Constants.OS_WEB    : return new WebTargetConfiguration(paths, config);
+            case Constants.OS_LINUX  : return new LinuxTargetConfiguration(paths, config, javaVersion);
+            case Constants.OS_DARWIN : return new MacOSTargetConfiguration(paths, config, javaVersion);
+            case Constants.OS_WINDOWS: return new WindowsTargetConfiguration(paths, config, javaVersion);
+            case Constants.OS_IOS    : return new IosTargetConfiguration(paths, config, javaVersion);
+            case Constants.OS_ANDROID: return new AndroidTargetConfiguration(paths, config, javaVersion);
+            case Constants.OS_WEB    : return new WebTargetConfiguration(paths, config, javaVersion);
             default                  : return null;
         }
     }
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -147,6 +147,19 @@ public class InternalProjectConfiguration {
         return new Version(m.group(1));
     }
 
+    /**
+     * Returns the Java version that GraalVM bundles
+     * @return the Java version of the GraalVM build
+     */
+    public Version getJavaVersion() {
+        return javaVersion;
+    }
+
+    /**
+     * Returns the GraalVM internal version.
+     * This version corresponds with the version used for the GraalVM maven artifacts.
+     * @return The GraalVM internal version.
+     */
     public Version getGraalVersion() {
         return graalVersion;
     }
@@ -670,7 +683,7 @@ public class InternalProjectConfiguration {
     }
 
     /**
-     * try to find the Java version that GraalVM bundles
+     * Try to find the Java version that GraalVM bundles
      * @return the Java version of the GraalVM's build
      * @throws NullPointerException when the configuration is null
      * @throws IllegalArgumentException when the configuration doesn't contain a property graalPath
@@ -700,20 +713,14 @@ public class InternalProjectConfiguration {
         return new Version(m.group(1));
     }
 
-    /**
-     * Gets the Java version that GraalVM bundles
-     * @return the Java version of the GraalVM build
-     */
-    public Version getJavaVersion() {
-        return javaVersion;
-    }
-
     @Override
     public String toString() {
         return "ProjectConfiguration{" +
                 "graalPath='" + publicConfig.getGraalPath() + '\'' +
                 ", javaStaticSdkVersion='" + getJavaStaticSdkVersion() + '\'' +
                 ", javafxStaticSdkVersion='" + getJavafxStaticSdkVersion() + '\'' +
+                ", javaVersion=" + javaVersion +
+                ", graalVersion=" + graalVersion +
                 ", useJNI=" + useJNI +
                 ", useJavaFX=" + useJavaFX +
                 ", usePrismSW=" + usePrismSW +
--- a/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
@@ -66,7 +66,6 @@ public class InternalProjectConfiguration {
     private boolean useJavaFX = false;
     private boolean usePrismSW = false;
     private boolean enableCheckHash = true;
-    private boolean usesJDK11 = false;
     private boolean sharedLibrary = false;
     private boolean staticLibrary = false;
 
@@ -75,12 +74,14 @@ public class InternalProjectConfiguration {
     private List<String> releaseSymbolsList;
 
     private final ProjectConfiguration publicConfig;
+    private final Version javaVersion;
+    private final Version graalVersion;
 
     /**
      * Private projects configuration, which includes everything, including public settings
      * @param config public project configuration
      */
-    public InternalProjectConfiguration(ProjectConfiguration config) {
+    public InternalProjectConfiguration(ProjectConfiguration config) throws IOException {
 
         this.publicConfig = Objects.requireNonNull(config);
 
@@ -115,13 +116,19 @@ public class InternalProjectConfiguration {
         }
 
         performHostChecks();
+
+        javaVersion = findGraalVMJavaVersion();
+        graalVersion = findGraalVersion();
+        checkGraalVMJavaVersion();
+        checkGraalVMVersion();
+        checkGraalVMVendor();
     }
 
     public Path getGraalPath() {
         return Objects.requireNonNull(this.publicConfig.getGraalPath(), "GraalVM Path is not defined");
     }
 
-    public Version getGraalVersion() throws IOException {
+    private Version findGraalVersion() throws IOException {
         String pattern = "GraalVM .*?(\\d\\d.\\d.\\d)";
         ProcessRunner graalJava;
         try {
@@ -138,6 +145,10 @@ public class InternalProjectConfiguration {
         if (!m.find())
             throw new IOException("Couldn't determine GraalVM version");
         return new Version(m.group(1));
+    }
+
+    public Version getGraalVersion() {
+        return graalVersion;
     }
 
     /**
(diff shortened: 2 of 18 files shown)


src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
```diff
@@ -385,23 +385,23 @@ public SubstrateDispatcher(Path buildRoot, ProjectConfiguration config) throws I
 
         Triplet targetTriplet = config.getTargetTriplet();
 
-        this.targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet),
+        this.targetConfiguration = Objects.requireNonNull(getTargetConfiguration(targetTriplet, javaVersion),
```
jperedadnr, line 388: Since Java/GraalVM versions are determined (also after #1234) from `InternalProjectConfiguration`, don't you think it should be better to keep the `javaVersion` as a field there, and then add a getter?
That would definitely prevent from needing to pass it all over the place.

In fact, we already have `usesJDK11`. Probably better to remove this one and just have `javaVersion` that can be queried from anywhere.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -74,8 +74,6 @@ public abstract class AbstractTargetConfiguration implements TargetConfiguration
             "png", "jpg", "jpeg", "gif", "bmp", "ttf", "raw",
             "xml", "fxml", "css", "gls", "json", "dat",
             "license", "frag", "vert", "obj", "mtl", "js");
-    protected static final List<String> ENABLED_FEATURES =
```
jperedadnr, line 77: Keep the feature as a static String, like `HOME_FINDER_FEATURE`, and add a comment like valid until GraalVM 21 or so.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -492,14 +493,23 @@ private String getNativeImagePath() {
...
-                .map(feature -> "--features=" + feature)
-                .collect(Collectors.toList());
+    protected List<String> getEnabledFeatures() {
+        return List.of();
+    }
+
+    private List<String> getEnabledFeaturesArgs() {
+        List<String> args = new ArrayList<>();
+        if (javaVersion.getMajor() < 21) {
+            args.add("--features=org.graalvm.home.HomeFinderFeature");
```
jperedadnr, line 503: `args.add("--features=" + HOME_FINDER_FEATURE);`

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -85,7 +83,8 @@ public abstract class AbstractTargetConfiguration implements TargetConfiguration
             "-H:+ReportExceptionStackTraces",
             "-H:-DeadlockWatchdogExitOnTimeout",
             "-H:DeadlockWatchdogInterval=0",
-            "-H:+RemoveSaturatedTypeFlows"
+            "-H:+RemoveSaturatedTypeFlows",
+            "--add-exports=org.graalvm.nativeimage.builder/com.oracle.svm.core.jdk=ALL-UNNAMED" // Required for GluonFeature
```
jperedadnr, line 87: GluonFeature is only used by the Linux target. You'll need to apply this argument only in that case, so I'd suggest adding another method like `getNativeImageArguments()` and override it from Linux.

src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
```diff
@@ -63,6 +63,13 @@ public class LinuxTargetConfiguration extends PosixTargetConfiguration {
             "java", "nio", "zip", "net", "prefs", "j2pkcs11", "sunec", "extnet", "fdlibm",
             "fontmanager", "javajpeg", "lcms", "awt_headless", "awt"
     );
+    /**
+     * fdlibm no longer required, unsure since when
```
jperedadnr, line 67: `fdlibm` is also no longer required on MacOSTarget.
Therefore, we should probably keep one single `staticJavaLibs` per target with it, and then in `getTargetSpecificJavaLinkLibraries()` add another filter for it, something like:
```
private List<String> getTargetSpecificJavaLinkLibraries() {
        return Stream.concat(getStaticJavaLibs().stream(), getOtherStaticLibs().stream())
                .filter(lib -> {
                    if (projectConfiguration.usesJDK11()) {
                        return true;
                    } else if (javaVersion.getMajor() < 21) {
                        return !lib.contains("sunec");
                    }
                    return !lib.contains("sunec") && !lib.contains("fdlibm");
                })
                .map(this::getLinkLibraryOption)
                .collect(Collectors.toList());
    }
```

jperedadnr (commented):
I've left some comments. 
Tested successfully (with some changes) on Mac with GraalVM 22.1.0.1-Final and GraalVM 21+35.1 (using also #1234)

src/main/java/com/gluonhq/substrate/SubstrateDispatcher.java
kkriske: To do that cleanly, I'd have to execute the version parsing logic in the constructor of the `InternalProjectConfiguration`. Would that be acceptable? I'm not sure what the reason is to delay it at the moment.

jperedadnr: That's fine, so far we do in `SubstrateDispatcher`:

```
        this.config = new InternalProjectConfiguration(config);
        Version javaVersion = this.config.checkGraalVMJavaVersion();
        this.config.checkGraalVMVersion(javaVersion);
        this.config.checkGraalVMVendor();
```

so all of that can go into the constructor, indeed (which already does for instance `performHostChecks`)

src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
kkriske: I'm not sure it's that straightforward.
For Window, both `mswsock` in the "other libs" and `extnet` in the "java libs" are new in the list for GraalVM for JDK 21.
I guess it's possible in your setup if for every future addition, there is a matching filter for every version below that, but I think that'll quickly get quite ugly and requires 2 separate modifications to add a library.

An alternative would be to build the list additively in the `getStaticJavaLibs` and `getOtherStaticLibs` methods e.g.
```java
    List<String> getStaticJavaLibs() {
        int major = projectConfiguration.getJavaVersion().getMajor();
        List<String> libs = new ArrayList<>(List.of(
                "j2pkcs11", "java", "net", "nio", "prefs", "zip", "sunmscapi"));
        if (major == 11) {
            libs.add("sunec");
        }
        if (major < 21) {
            libs.add("fdlibm");
        }
        if (major >= 21) {
            libs.add("extnet");
        }
        return libs;
    }
```

But that does away with the static list that's currently used.

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -140,6 +147,10 @@ public Version getGraalVersion() throws IOException {
         return new Version(m.group(1));
     }
 
+    public Version getGraalVersion() {
```
jperedadnr, line 150: add javadoc

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -663,13 +670,13 @@ private void checkGraalVMPermissions(String graalvmHome) {
     }
 
     /**
-     * Gets the Java version that GraalVM bundles
+     * try to find the Java version that GraalVM bundles
```
jperedadnr, line 673: Minor `s/try/Try`

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -693,6 +700,14 @@ public Version getGraalVMJavaVersion() throws IOException {
         return new Version(m.group(1));
     }
 
+    /**
```
jperedadnr, line 703: move this up, right before `getGraalVersion()`

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -693,6 +700,14 @@ public Version getGraalVMJavaVersion() throws IOException {
...
+     * Gets the Java version that GraalVM bundles
+     * @return the Java version of the GraalVM build
+     */
+    public Version getJavaVersion() {
+        return javaVersion;
+    }
+
```
jperedadnr, line 710: add java/graalVersion to `toString()`

src/main/java/com/gluonhq/substrate/target/LinuxTargetConfiguration.java
jperedadnr: Right, I missed the `extnet` addition.

We want to have the static libs declared on top, but I like the change you propose, so what about (for Windows, that is):

```
    private static final List<String> staticJavaLibsBase = Arrays.asList(
            "j2pkcs11", "java", "net", "nio", "prefs", "zip", "sunmscapi");
    private static final List<String> staticJavaLibs11 = List.of("sunec");
    private static final List<String> staticJavaLibs20 = List.of("fdlibm");
    private static final List<String> staticJavaLibs21 = List.of("extnet");


@Override
    List<String> getStaticJavaLibs() {
        int major = projectConfiguration.getJavaVersion().getMajor();
        List<String> libs = staticJavaLibsBase;
        if (major == 11) {
            libs.addAll(staticJavaLibs11);
        }
        if (major < 21) {
            libs.addAll(staticJavaLibs20);
        }
        if (major >= 21) {
            libs.addAll(staticJavaLibs21);
        }
        return libs;
    }
```

If you do it like this in the other targets, we could get rid of the filter in `getTargetSpecificJavaLinkLibraries` to remove `sunec`.
Also, don'f forget about `fdlibm` in macOS.

kkriske: I've appliled this in the `AbstractTargetConfiguration` class as well, assuming it's necessary for the targets outside of Linux/Windows/MacOS as well.
It's applied like this for now, one thing I don't like that much is the amount of duplication that's currently in there.

It would be nice having just a single list of libraries wrapped in a version range..
```java
List<JavaLib> staticJavaLibs = List.of(
        // applied to all versions
        JavaLib.of("java"),
        // applied up to specific version
        JavaLib.upTo(11, "sunec"),
        // applied from specific version
        JavaLib.from(21, "extnet"),
        // applied within version range, not currently applicable
        JavaLib.range(11, 21, "example")
    );
```
That's a single field with everything in it, after which it can easilly be filtered in the abstract parent as such:
```java
int major = projectConfiguration.getJavaVersion().getMajor();
staticJavaLibs.stream()
        .filter(l -> l.inRange(major))
        .map(JavaLib::getLibName)
        .toList();
```
I feel this is getting a bit off track though.

jperedadnr: yes, and on top of that, we are missing one important thing (sorry, my bad): the order of the libraries does matter. 

So we can't just remove them (fdlibm, sunec) from the initial ordered list, and add them at the end when they are needed, that won't work. 

That's why we had the filter to remove sunec in the first place, and what my initial proposal was about.

So I'd say we should go back to it:
```
private List<String> getTargetSpecificJavaLinkLibraries() {
       int major = projectConfiguration.getJavaVersion().getMajor();
       return Stream.concat(getStaticJavaLibs().stream(), getOtherStaticLibs().stream())
                .filter(lib -> {
                    if (major == 11) {
                        return true; // all are needed
                    } else if (major < 21) {
                        return !lib.contains("sunec"); // remove sunec, keep fdlibm
                    }
                    return !lib.contains("sunec") && !lib.contains("fdlibm"); // remove sunec, fdlibm
                })
                .map(this::getLinkLibraryOption)
                .collect(Collectors.toList());
    }
```

For Windows, you can simply add extnet (assuming this can go at the end of the list):
```
@Override
    List<String> getStaticJavaLibs() {
        List<String> libs = new ArrayList<>(staticJavaLibs);
        if (projectConfiguration.getJavaVersion().getMajor() >= 21) {
            libs.addAll(staticJavaLibs21);
        }return libs;
    }
```
Then linux and Mac don't need changes on this matter.

Sorry for the confusion, but does this make sense to you?

kkriske: What about the single list with version bounds? That would preserve order and IMO is more readable and maintainable than the filter.

jperedadnr: yes, that could work as well, of course, seems more complex to implement, but you are right about future maintenance.

src/main/java/com/gluonhq/substrate/util/JavaLib.java
```diff
@@ -0,0 +1,77 @@
...
+
+    public static JavaLib range(int lowerBound, int upperBound, String libName) {
+        return new JavaLib(lowerBound, upperBound, libName);
+    }
+
+    public static JavaLib from(int lowerBound, String libName) {
+        return new JavaLib(lowerBound, null, libName);
+    }
+
+    public static JavaLib upto(int upperBound, String libName) {
```
jperedadnr, line 74: `s/upto/upTo`

src/test/java/com/gluonhq/substrate/util/JavaLibTest.java
```diff
@@ -0,0 +1,67 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package com.gluonhq.substrate.util;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class JavaLibTest {
```
jperedadnr, line 34: same here, rename to `LibTest`

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -98,8 +101,21 @@ public abstract class AbstractTargetConfiguration implements TargetConfiguration
     protected final boolean crossCompile;
 
     private final List<String> defaultAdditionalSourceFiles = Collections.singletonList("launcher.c");
-    private final List<String> defaultStaticJavaLibs = List.of("java", "nio", "zip", "net", "prefs", "jvm",
-            "fdlibm", "z", "dl", "j2pkcs11", "sunec", "jaas", "extnet");
+    private final List<JavaLib> defaultStaticJavaLibs = List.of(
+            JavaLib.of("java"),
```
jperedadnr, line 105: minor, I'd pack them in a few lines, 4 per line or so.

src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -929,6 +955,19 @@ List<String> getOtherStaticLibs() {
         return List.of();
     }
 
+    /**
+     * Return the list of library names applicable to the used java version.
+     * @param libs List to validate based on {@link JavaLib#inRange(int)}.
+     * @return The list of library names applicable to the used java version.
+     */
+    List<String> filterApplicableLibs(List<JavaLib> libs) {
```
jperedadnr, line 963: might be `final`, as this is not meant to be overridden

jperedadnr:
@kkriske please fix conflicts (and do the refactoring if needed)

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
```diff
@@ -664,31 +674,30 @@ private void checkGraalVMPermissions(String graalvmHome) {
...
-        List<String> responses = graalJava.getResponses();
-        if (responses == null || responses.isEmpty()) {
-            throw new IOException("Couldn't determine GraalVM's Java version");
-        }
-        return parseGraalVMJavaVersion(responses.get(0));
     }
 
     static Version parseGraalVMJavaVersion(String versionString) {
-        String realVersion = versionString.replaceAll("-internal", "");
+        String realVersion = versionString.replace("-internal", "");
```
kkriske, line 700: I don't know exactly what case this `-internal` string covers, but it wouldn't hurt adding a matching parser test as well.

jperedadnr: If you clone the JDK repo and build it yourself you get something like:
```
$ jdk/build/macosx-aarch64-server-release/images/jdk/bin/java -version
openjdk version "22-internal" 2024-03-19
OpenJDK Runtime Environment (build 22-internal-adhoc.<$user.name>.jdk)
OpenJDK 64-Bit Server VM (build 22-internal-adhoc.<$user.name>.jdk, mixed mode, sharing)
```

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
jperedadnr: A similar case is if you run with an `ea` version:
```
openjdk version "22-ea" 2024-03-19
OpenJDK Runtime Environment (build 22-ea+24-1921)
OpenJDK 64-Bit Server VM (build 22-ea+24-1921, mixed mode, sharing)
```

src/main/java/com/gluonhq/substrate/model/InternalProjectConfiguration.java
jperedadnr: @kkriske Just for my understanding, are you planning on adding more tests to the internal (and maybe the "ea" case, if you use https://jdk.java.net/22/ for instance), or is this PR ready?

kkriske: What's still left to get this merged?
Should I add these -internal and -ea version string? Those are not supported I'd think because they don't contain any GraalVM version, which resullts in a `IllegalArgumentException`.

jperedadnr: For me it should be ready. 

I was just asking because of your comment:
> but it wouldn't hurt adding a matching parser test as well.

We probably don't need to worry about testing those cases now. `-internal` is removed, and GraalVM doesn't use `-ea` (just tested latest nightly build from [here](https://github.com/graalvm/graalvm-ce-dev-builds/releases/)).

kkriske: I missed your last message. I'm currently not planning on adding anything more, mainly because I don't really understand the use-cases of the ea and internal versions because the version checks will reject them as they do not have GraalVM versions.

kkriske: @jperedadnr Just wanna bump this.
If nothing is still blocking this, can we get this released?

abhinayagarwal (approved):

credmond:
> I've left some comments. 
> Tested successfully (with some changes) on Mac with Gluon's GraalVM 22.1.0.1-Final and the official GraalVM 21+35.1 (using also #1234) 
> 
> In any case, note that the official GraalVM 21 builds will still fail building native images for mobile targets (more changes are required on Substrate, but also probably on the GraalVM builds)

Hey...could you tell me if Gluon is planning on a new official Gluon GraalVM build anytime soon?

PR #1240, opened 2023-12-20
Make ie4uinit.exe execution optional

### LINKED ISSUE #1239: Build on WindowsServerCore container fails
Build fails on WindowsServerCore container, because `ie4uinit` dos not exist. I have not yet found a way to install this into the container as it doesn't have a Desktop UI, and ie4uinit is part of UI related stuff.

Fix would be to check if ie4uinit exists, and not run it if it doesn't.
Alternatively, ignore exceptions if it doesn't run, as it's only cosmetic.

### Expected Behavior

Build succeeds

### Current Behavior

This piece of code crashes:
https://github.com/gluonhq/substrate/blob/bd0f52be101a7e2ad3e237f2288855e59d307ca6/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java#L287-L291

Because `ie4uinit` does not exist.

### Steps to Reproduce

Do a build on WindowsServerCore.

### Your Environment
WindowsServerCore 2022 container, substrate 0.0.61

kkriske:
- Do not run ie4uinit if it is not on the PATH.
- Uses 'where' to check if an executable is on the PATH

### Issue

Build on WindowsServerCore containers fail because `ie4uinit` does not exist.

Fixes #1239

### Progress

- [x] Verify the contributor has signed [Gluon Individual Contributor License Agreement (CLA)](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
@@ -285,6 +285,7 @@ public class WindowsTargetConfiguration extends AbstractTargetConfiguration {
     // During development if user changes the application icon, the same is not reflected immediately in Explorer.
     // To fix this, a cache clearance of the Windows explorer is required.
     private void clearExplorerCache() throws IOException, InterruptedException {
+        if (!executableOnPath("ie4uinit.exe")) return;
         ProcessRunner clearCache = new ProcessRunner("ie4uinit");
         clearCache.addArg(findCacheFlag());
         clearCache.runProcess("Clear Explorer cache");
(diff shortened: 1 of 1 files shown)


src/main/java/com/gluonhq/substrate/target/WindowsTargetConfiguration.java
```diff
@@ -285,6 +285,7 @@ private void createIconResource() throws InterruptedException, IOException {
     // During development if user changes the application icon, the same is not reflected immediately in Explorer.
     // To fix this, a cache clearance of the Windows explorer is required.
     private void clearExplorerCache() throws IOException, InterruptedException {
+        if (!executableOnPath("ie4uinit.exe")) return;
```
jperedadnr, line 288: Add a log if this happens, warning the user that the application icon might have been cached and not updated properly?

jperedadnr (approved):

PR #1273, opened 2024-09-04
Update Substrate to use latest GraalVM (JDK 23) and JavaFX 24-ea

jperedadnr:
### Issue

Fixes #1241 

### Progress

- [ ] Change must not contain extraneous whitespace
- [ ] License header year is updated, if required
- [ ] Verify the contributor has signed [Gluon Individual Contributor License Agreement (CLA)](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)

PR #1278, opened 2024-09-08
checks for static lib

jperedadnr:
After #1273, add some fixes for `staticlib()`, and remove unneeded argument.


src/main/java/com/gluonhq/substrate/target/AbstractTargetConfiguration.java
```diff
@@ -81,7 +81,6 @@ public abstract class AbstractTargetConfiguration implements TargetConfiguration
 
     private static final List<String> baseNativeImageArguments = Arrays.asList(
             "-Djdk.internal.lambda.eagerlyInitialize=false",
-            "-Ddebug.jdk.graal.jvmciConfigCheck=warn",
```
johanvos, line 84: shouldn't this be defined somewhere else then? (globally?)

jperedadnr: it is not needed once https://github.com/gluonhq/graal/pull/20 gets merged

johanvos (approved):

PR #1282, opened 2024-09-11
Bump static JDK version

jperedadnr:
Bump vmone version to gvm-24-2

### DIFF
--- a/src/main/resources/native/ios/JvmFuncsFallbacks.c
+++ b/src/main/resources/native/ios/JvmFuncsFallbacks.c
@@ -28,11 +28,6 @@ JNIEXPORT jobject JNICALL JVM_AreNestMates(JNIEnv *env) {
     return NULL;
 }
 
-JNIEXPORT jobject JNICALL JVM_ArrayCopy(JNIEnv *env) {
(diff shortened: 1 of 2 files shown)


src/main/resources/native/ios/JvmFuncsFallbacks.c
```diff
@@ -28,11 +28,6 @@ JNIEXPORT jobject JNICALL JVM_AreNestMates(JNIEnv *env) {
     return NULL;
 }
 
-JNIEXPORT jobject JNICALL JVM_ArrayCopy(JNIEnv *env) {
```
johanvos, line 31: As a follow-up, this whole file should be removed, as it is obsoleted by libvmone.

johanvos (approved):

src/main/resources/native/ios/JvmFuncsFallbacks.c
jperedadnr: I take that all these dummy functions are not used at all, but are not in vmone yet (otherwise we'd get duplicated symbols). Do you mean we should move them to vmone?

PR #1319, opened 2025-04-20
Updated the clibPath to point to the glibc directory that contains th…

### LINKED ISSUE #1318: Linux x86_64 build failure with GraalVM 23.0.2
Gluon Maven Plugin fails to build executable on linux x86_64 with Oracle GraalVM 23.0.2 with the following error:
```
java.io.IOException: Missing library libjvm.a not in linkpath /home/gillman/.sdkman/candidates/java/23.0.2-graal/lib/svm/clibraries/linux-amd64
	at com.gluonhq.substrate.target.PosixTargetConfiguration.checkPlatformSpecificClibs(PosixTargetConfiguration.java:47)
	at com.gluonhq.substrate.target.AbstractTargetConfiguration.ensureClibs(AbstractTargetConfiguration.java:454)
	at com.gluonhq.substrate.target.AbstractTargetConfiguration.link(AbstractTargetConfiguration.java:199)
	at com.gluonhq.substrate.target.LinuxTargetConfiguration.link(LinuxTargetConfiguration.java:135)
	at com.gluonhq.substrate.SubstrateDispatcher.nativeLink(SubstrateDispatcher.java:440)
	at com.gluonhq.NativeLinkMojo.execute(NativeLinkMojo.java:47)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)
```

The file appears to be in a subdirectory, either glibc or musl.

Also appears to be an issue in this older ticket: https://github.com/gluonhq/gluon-samples/issues/119

### Expected Behavior
Executable should be built.

### Current Behavior
Executable build fails.

### Steps to Reproduce
Install Oracle GraalVM and then build any project using the gluon maven plugin.

### Your Environment
Linux x86_64
Oracle GraalVM 23.0.2

wynan:
Updated the clibPath in the Posix Configuration based on the information [here](https://www.graalvm.org/latest/reference-manual/native-image/guides/build-static-executables/) and based on the location of my local files.
Updated the build.gradle file to address the gradle warnings.

### Issue

Fixes #1318

### Progress

- [X] Change must not contain extraneous whitespace
- [X] License header year is updated, if required
- [X] Verify the contributor has signed [Gluon Individual Contributor License Agreement (CLA)](https://docs.google.com/forms/d/16aoFTmzs8lZTfiyrEm8YgMqMYaGQl0J8wA0VJE2LCCY)

### DIFF
--- a/src/main/java/com/gluonhq/substrate/target/PosixTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/PosixTargetConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, 2022, Gluon
+ * Copyright (c) 2019, 2022, 2025, Gluon
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/main/java/com/gluonhq/substrate/target/PosixTargetConfiguration.java
+++ b/src/main/java/com/gluonhq/substrate/target/PosixTargetConfiguration.java
@@ -42,6 +42,11 @@ abstract class PosixTargetConfiguration extends AbstractTargetConfiguration {
     }
 
     @Override
+    protected Path getCLibPath() {
+        return super.getCLibPath().resolve("glibc");
+    }
+
+    @Override
     void checkPlatformSpecificClibs(Path clibPath) throws IOException {
         Path libjvmPath = clibPath.resolve("libjvm.a");
         if (!Files.exists(libjvmPath)) throw new IOException("Missing library libjvm.a not in linkpath "+clibPath);
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,9 @@ plugins {
 
 group = "com.gluonhq"
 
-sourceCompatibility = JavaVersion.VERSION_11
+java {
+    sourceCompatibility = JavaVersion.VERSION_11
+}
 
 repositories {
     mavenCentral()


ctoabidmaqbool:
@jperedadnr @johanvos @abhinayagarwal

src/main/java/com/gluonhq/substrate/target/PosixTargetConfiguration.java
```diff
@@ -41,6 +41,11 @@ abstract class PosixTargetConfiguration extends AbstractTargetConfiguration {
         super(paths, configuration);
     }
 
+    @Override
+    protected Path getCLibPath() {
+        return super.getCLibPath().resolve("glibc");
```
jperedadnr, line 46: Currently, the GluonFX 1.0.26 version is meant to be used with GraalVM 22.1.0.1-Final (JDK 17 based, from here https://github.com/gluonhq/graal/releases/), but nothing prevents you from using any other higher GraalVM version.

However, this is not recommended, as that in fact it will surely fail on Android/iOS (for instance, the CAP cache files change with new GraalVM versions, so the current ones in Substrate are guaranteed to work for a given GraalVM version).

The previous GluonFX 1.0.25, non-supported/experimental, version was meant to be used with GraalVM 23+25.1-dev, but that one is not failing with the error you have posted, because it doesn't use clibs.

So the error only happens in the unexpected combination of GluonFX 1.0.26 + GraalVM 23 + Linux x86_64. In any case, it doesn't happen on macOS.

If you want to move forward with this PR, I suggest you move this change to the LinuxTargetConfiguration class, and then you can check if the directory `glibc` exists before adding it to the path.

